PR #14, opened 2015-04-29
move spec document to root

cvan:
this would let folks link to https://w3c.github.io/gamepad/ instead of https://w3c.github.io/gamepad/gamepad.html


sgraham:
That seems better, but are there existing links to the old location? Maybe a stub there?

cvan:
ah, very good point. I've changed `gamepad.html` to now contain only code to handle these redirects:

https://w3c.github.io/gamepad/gamepad.html#remapping → https://w3c.github.io/gamepad/#remapping
`file:///opt/gamepad/gamepad.html#remapping` → `file:///opt/gamepad/index.html#remapping`

AFBarstow:
@sgraham the version at w3.org/TR/gamepad/ references gamepad.html but updating the TR version is trivial since the spec uses ReSpec and I can do the TR publishing using the "echidna". My recommendation is to make the name change now and then I'll immediately publish a new WD.

sgraham:
Thanks

AFBarstow:
@sgraham FYI, a new TR of gamepad was just published w3.org/TR/gamepad/ (symlink'd to http://www.w3.org/TR/2015/WD-gamepad-20150429/).

(Not sure if gamepad.html is needed anymore.)

cvan:
> (Not sure if gamepad.html is needed anymore.)

as long as folks weren't linking to [it](https://w3c.github.io/gamepad/gamepad.html), yeah. [Google shows](https://www.google.com/webhp#q="https%3A%2F%2Fw3c.github.io%2Fgamepad%2Fgamepad.html") only a single link from a mailing-list thread.

Issue #4, opened 2015-04-09
Should fire events instead of using passive model


AshleyScirra:
(This is based on an email sent to public-webapps)

The Gamepad API is the only web platform input API that relies on passively polling the input state. Gamepads should fire events like "gamepadbuttondown", "gamepadbuttonup" and "gamepadaxismove". This would have the following benefits:
- consistent with other input APIs (e.g. "mousedown", "mouseup", "mousemove" / "touchstart", "touchmove", "touchend" / "pointerdown", "pointermove", "pointerup")
- easier to program for: if you want a "gamepadbuttondown" event, you have to implement it yourself by polling the state, keeping the previous polled state, and comparing the differences looking for a previously up but currently down state and then run your handler
- easily integrates with browser's existing "user gesture" model for sensitive features like opening popup windows, copying to the clipboard, and playing audio or video on mobile. Since gamepads do not fire events, currently it is not possible to use any "user gesture" features with gamepad input alone in any existing browsers.

pyalot:
Note that using events exlusively without control over the event loop has severe data handling drawbacks.

If events are introduced, they should be issued at a controlled point in time (when polling) which clears the entire event queue up to that point, and they should not substitute "passive" (it's not really passive, it's active polling) querying of values, particularly not axis values.

luser:
/cc @sgraham @toji

This has been discussed on public-webapps in the past:
https://lists.w3.org/Archives/Public/public-webapps/2012AprJun/0469.html

and it's on my list of features for a next version of the spec:
https://www.w3.org/wiki/Webapps/GamepadFeatures

Firefox includes non-standard events that you can enable by flipping a pref (dom.gamepad.non_standard_events.enabled), you'll get GamepadButton{Down,Up} and GamepadAxisMove events.

I think spec'ing the button events would be fairly simple and uncontroversial and solves the most important use case--missing button presses that happen between polls.

Events for axis values (or analog button values) are harder to get right. They're likely to be spammy and not incredibly useful--most consumers just want to know the current value of an axis. It would be nice to have _some_ sort of event for this so that consumers aren't forced to constantly poll. Maybe we could write some spec text such that we'd fire a "datachanged" event whenever anything changes, but the browser can roll up a series of them into a single event, like if you had:

```
<button 1 pressed>
<axis 1 moved>
<axis 2 moved>
<axis 1 moved>
<axis 2 moved>
<button 1 released>
<axis 1 moved>
```

the browser could compress this to:

```
[...]

Issue #8, opened 2015-04-24
 Need to spec liveness of Gamepad objects


luser:
Originally filed as https://www.w3.org/Bugs/Public/show_bug.cgi?id=21434:

 Ted Mielczarek [:ted] 2013-03-29 13:34:49 UTC

The discussion in bug 17309 made it clear that Scott and I have implemented two different things in regards to the "liveness" of Gamepad objects. Scott's Chrome implementation returns a static snapshot of each Gamepad when you getGamepads(), whereas the Gamepad object attached to the "ongamepadconnected" event in my Firefox implementation is live: you can get the latest device state out of it at any time.

We need to decide what the correct behavior here is and spec it.

Obviously I'm biased in favor of my implementation, but I think it does have one large benefit: we're expecting users to poll Gamepad status in a requestAnimationFrame loop, so if you're generating new Gamepad objects for every call to getGamepads() you're going to be generating a lot of garbage.

[reply] [−] Comment 1 Scott Graham 2013-03-29 17:26:04 UTC

(In reply to comment #0)

> The discussion in bug 17309 made it clear that Scott and I have implemented
> two different things in regards to the "liveness" of Gamepad objects.
> Scott's Chrome implementation returns a static snapshot of each Gamepad when
> you getGamepads(), whereas the Gamepad object attached to the
> "ongamepadconnected" event in my Firefox implementation is live: you can get
> the latest device state out of it at any time.
> 
> We need to decide what the correct behavior here is and spec it.
> 
> Obviously I'm biased in favor of my implementation, but I think it does have
> one large benefit: we're expecting users to poll Gamepad status in a
> requestAnimationFrame loop, so if you're generating new Gamepad objects for
> every call to getGamepads() you're going to be generating a lot of garbage.

(I'm honestly not sure precisely what I implemented.)

I agree on the reducing garbage concern, and so returning/modifying the same objects sounds right to me.

The part I don't like as much is that the data unexpectedly changes, and it's unclear when, and with what frequency it's updated. That's why I used getGamepads() as the sync point for "Please update the data in my gamepad objects to the most recent state".

[reply] [−] Comment 2 Ted Mielczarek [:ted] 2013-04-17 13:14:21 UTC

The HTML WebApp spec has concepts we can use to define this in a straightforward way:
http://www.whatwg.org/specs/web-apps/current-work/multipage/webappapis.html#concept-task

It basically specs thread event queues, so that you can say that the Gamepad state will be updated after the current script has finished executing.

[...]

PR #15, opened 2015-04-29
spec `gamepadbuttondown` and `gamepadbuttonup` events (issue #4)

cvan:
This is a recommendation to spec the button events that Firefox has implemented: `gamepadbuttondown` and `gamepadbuttonup` (there's one other non-standard event, `gamepadaxismove`, which I've omitted for now).

Per @luser in https://github.com/w3c/gamepad/issues/4#issuecomment-95950556:

> Firefox includes non-standard events that you can enable by flipping a pref (dom.gamepad.non_standard_events.enabled), you'll get GamepadButton{Down,Up} and GamepadAxisMove events.
> 
> I think spec'ing the button events would be fairly simple and uncontroversial and solves the most important use case--missing button presses that happen between polls.

It's worth noting that this PR doesn't address the other issues outlined in issue #4: whether to introduce axis and/or button change events (à la `keypress` vs. `keydown`+`keyup`).

This is my first contribution, so apologies in advance if this isn't up to snuff. I'm all ears. Thanks for taking a look!

P.S. I have a local branch where I've grouped the interfaces and events into their own respective sections, so the ToC reads a bit more hierarchically. I can submit later, if folks think it's warranted.
### Relevant links
- [code in Firefox for nonstandard events](https://dxr.mozilla.org/mozilla-central/source/dom/gamepad/GamepadService.cpp)
- [a simple demo page with listeners for all the current gamepad events](http://jsbin.com/recudojega/1/) (enable `dom.gamepad.non_standard_events.enabled` in `about:config`)
- [a library that polyfills all the events](https://github.com/cvan/gamepad-plus/blob/master/main.js) where unsupported

### DIFF
--- a/index.html
+++ b/index.html
@@ -442,6 +442,171 @@
     </section>
 
     <section>
+      <h2><dfn>GamepadButtonEvent</dfn> Interface</h2>
+      <dl title='[Constructor(GamepadButtonEventInit eventInitDict)] interface GamepadButtonEvent : Event'
+          class='idl'>
+        <dt>readonly attribute Gamepad gamepad</dt>
+
+        <dd>
+
+            The single gamepad attribute provides access to the associated
+            gamepad data for this event.
+
+        </dd>
+
+        <dt>readonly attribute GamepadButton button</dt>
+
+        <dd>
+
+            The single button attribute provides access to the associated
+            button data for this event.
+
+        </dd>
+
+      </dl>
+      <dl title='dictionary GamepadButtonEventInit : EventInit' class='idl'>
+        <dt>required Gamepad gamepad</dt>
+        <dd>
+          The gamepad associated with this event.
+        </dd>
+        <dt>required GamepadButton button</dt>
+        <dd>
+          The button associated with this event.
+        </dd>
+      </dl>
+    </section>
+
+    <section>
+
+      <h3 id="event-gamepadconnected">The <dfn class="event">gamepadconnected</dfn> event</h3>
+
+        <p>
+          User agents implementing this specification must provide a new DOM
+          event, named <a><code>gamepadconnected</code></a>. The corresponding
+          event MUST be of type
+          <a href="#gamepadevent-interface"><code>GamepadEvent</code></a> and
+          MUST fire on the <code>window</code> object. Registration for and
+          firing of the <a><code>gamepadconnected</code></a> event MUST follow
+          the usual behavior of DOM4 Events. [[!DOM4]]
+        </p>
+
+        <p>
+          A <a>user agent</a> MUST dispatch this event type to indicate the
+          user has connected a gamepad. If a gamepad was already connected
+          when the page was loaded, the <a><code>gamepadconnected</code></a>
+          event SHOULD be dispatched when the user presses a button or moves
+          an axis.
+        </p>
+
+    </section>
+
+    <section>
+
+        <h3 id="event-gamepaddisconnected">The <dfn class="event">gamepaddisconnected</dfn> event</h3>
+
+        <p>
+          User agents implementing this specification must provide a new DOM
+          event, named <a><code>gamepaddisconnected</code></a>. The
+          corresponding event MUST be of type
+          <a href="#gamepadevent-interface"><code>GamepadEvent</code></a>
+          and MUST fire on the <code>window</code> object. Registration for
+          and firing of the <a><code>gamepaddisconnected</code></a> event MUST
+          follow the usual behavior of DOM4 Events. [[!DOM4]]
+        </p>
+
+        <p>
+          When a gamepad is disconnected from the <a>user agent</a>, if the
+          <a>user agent</a> has previously dispatched a
+          <a href="event-gamepadconnected"><code>gamepadconnected</code></a>
+          event for that gamepad to a window, a
+          <a><code>gamepaddisconnected</code></a> event MUST be dispatched to
+          that same window.
+        </p>
+
+    </section>
+
+    <section>
+
+        <h3 id="event-gamepadbuttondown">The <dfn class="event">gamepadbuttondown</dfn> event</h3>
+
+        <p>
+          User agents implementing this specification must provide a new DOM
+          event, named <a><code>gamepadbuttondown</code></a>. The
+          corresponding event MUST be of type
+          <a href="#gamepadevent-interface"><code>GamepadEvent</code></a>
+          and MUST fire on the <code>window</code> object. Registration for
+          and firing of the <a><code>gamepadbuttondown</code></a> event MUST
+          follow the usual behavior of DOM4 Events. [[!DOM4]]
+        </p>
+
+        <p>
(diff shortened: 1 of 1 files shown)


cvan:
this PR has been open for quite a while. shall we close until we've discussed the pros and cons of polling vs. events?

this has recently come up in https://github.com/w3c/gamepad/issues/37#issue-177528094 and issue #22

nondebug:
> solves the most important use case--missing button presses that happen between polls.

Gamepad button (and axis) events are also important for applications that do not process inputs locally; i.e., when button and axis state is streamed to a remote service that manages the application state, waiting until rAF time to handle inputs adds a significant amount of input latency. Firing events for these inputs would allow the state to be streamed more quickly after the device is polled.

nondebug:
> gamepadbuttondown and gamepadbuttonup

This assumes that gamepad buttons are digital, but many devices expose buttons with analog values that can change without toggling the button's .pressed state. For instance, analog triggers on Xbox 360 or pressure-sensitive buttons on Dualshock 3. I think a gamepadbuttonmove event would also make sense here, in addition to the down/up transition events.

index.html
```diff
@@ -441,6 +441,171 @@ <h2><dfn>GamepadEvent</dfn> Interface</h2>
...
+    </section>
+
+    <section>
+
+        <h3>Other events</h3>
+
+        <p>
+
+          <i>More discussion is needed on whether to include events for axis
+          changes (e.g., <code>gamepadaxismove</code> for whenever the value of
```
nondebug, line 601: Please also consider adding a gamepadbuttonmove event in order to better support gamepads that expose buttons with analog values.

index.html
```diff
@@ -441,6 +441,171 @@ <h2><dfn>GamepadEvent</dfn> Interface</h2>
...
+        <dt>readonly attribute Gamepad gamepad</dt>
+
+        <dd>
+
+            The single gamepad attribute provides access to the associated
+            gamepad data for this event.
+
+        </dd>
+
+        <dt>readonly attribute GamepadButton button</dt>
```
nondebug, line 457: Firefox's implementation uses a button index instead of a GamepadButton:

gamepadbuttondown { target: Window → /, isTrusted: true, **button: 1**, gamepad: Gamepad, currentTarget: Window → /, eventPhase: 2, bubbles: false, cancelable: false, defaultPrevented: false, composed: false, timeStamp: 1530125797442210 }

I think it would be better to include the button index to make it easier for listeners to know which button was pressed. The GamepadButton can be accessed through the gamepad attribute once we know the index, but it's difficult to determine a button's index from a GamepadButton.

index.html
```diff
@@ -441,6 +441,171 @@ <h2><dfn>GamepadEvent</dfn> Interface</h2>
...
+
+        <dd>
+
+            The single button attribute provides access to the associated
```
nondebug, line 461: The event is asynchronous, but gamepad data is polled synchronously. This means that the button state in GamepadButton may not match the state as described by the event. For instance, for a gamepadbuttondown event, e.button.pressed may no longer be true if the button was released between the time the event was fired and when it was handled.

This proposal should be modified to indicate if the state in e.button is intended to preserve the button's state when the event was generated or if it should report the current button state.

index.html
```diff
@@ -441,6 +441,171 @@ <h2><dfn>GamepadEvent</dfn> Interface</h2>
...
+    <section>
+
+        <h3>Other events</h3>
+
+        <p>
+
+          <i>More discussion is needed on whether to include events for axis
+          changes (e.g., <code>gamepadaxismove</code> for whenever the value of
+          an axis changes, or <code>gamepadaxischange</code> for whenever the
+          active axis changes), etc.</i>
```
nondebug, line 603: What is an active axis?

index.html
```diff
@@ -441,6 +441,171 @@ <h2><dfn>GamepadEvent</dfn> Interface</h2>
...
+
+    <section>
+
+        <h3 id="event-gamepadbuttondown">The <dfn class="event">gamepadbuttondown</dfn> event</h3>
+
+        <p>
+          User agents implementing this specification must provide a new DOM
+          event, named <a><code>gamepadbuttondown</code></a>. The
+          corresponding event MUST be of type
+          <a href="#gamepadevent-interface"><code>GamepadEvent</code></a>
```
nondebug, line 536: GamepadButtonEvent

index.html
cvan: good call, thanks!

index.html
cvan: good call, will remove the part about `active`; that doesn't make sense. multiple axes could change. perhaps "axis changes" could be detected by higher-level abstractions (content libraries).

marcoscaceres:
Is this still a thing that's going to be added to the spec?

cvan:
> Is this still a thing that's going to be added to the spec?

I think there's still interest from web developers in having an event model for button changes (or an improved API for polling).

related: see @nondebug's comment here: https://github.com/w3c/gamepad/issues/83#issuecomment-413636028

also see these various issues around polling:

#77 – Fast button presses could get lost
#46 – Arbitrary buffering of API calls is causing issues
#22 – Should spec a buffering/refresh interval model of gamepad data?
#8 – Need to spec liveness of Gamepad objects
#4 – Should fire events instead of using passive model

to answer your question: I hope so. does anyone disagree with the changes proposed in this PR? if so, which API changes might we be able to make so as to sanely surface button (and axis) changes?

mrmcpowned:
Has there been any more movement on this? I also think it'd be welcoming to have an event-driven model, but only as an inclusion alongside the polling model as well.

The following question was asked in #22:
> With that in mind, I guess my question to you is: do you know any practical examples where lost gamepad events are actually a real problem?

Lost gamepad events can be a real problem when the intended use is for accurate input visualization. A lost frame of data can be reasonable in terms of displaying said data, but an inconsistency between axis values or when two or more buttons were pressed simultaneously is not. 

The suggested approach to #4 to remedy such a scenario (the API providing an event subscription strategy for multiple axes/buttons/etc) doesn't account for this kind of scenario because it could create a disconnect between axes and button represented at a fixed point in time due to the asynchronicity.

We could then propose something like `onGamepadChange` that would subscribe to the entire gamepad, but something like that would feel like a bastardization of the polling strategy, and would that really be better than polling from the start?

It's why I'm not against the inclusion of events, but the removal of polling wouldn't allow for the kind of scenarios I'm depicting. For something like adding gamepads as a form of navigational input for a webpage, events would be the perfect approach since it'd be idiomatic in comparison to other event-driven practices and wouldn't necessarily encounter the kind of issues I mentioned above.

This also isn't a theoretical application either. I run and manage [gamepadviewer.com](gamepadviewer.com), which does exactly what it says, and the amount of usage is not insignificant, either. According to the [`getGamepads()` feature data](https://www.chromestatus.com/metrics/feature/timeline/popularity/1916), the site ranks at least 34th in the list, so everything I've posited isn't hyperbole.

marcoscaceres:
closing due to inactivity. We might revive this later.

Issue #1, opened 2015-04-09
UA applied mapping cannot be escaped


pyalot:
It's sometimes desirable not to use the UAs mapping (because it might be wrong, hide some controls or a user implemented handling provides better utility).

The current specification cannot escape a UA provided mapping. I'd suggest adding a way to escape mapping.

luser:
@kg and I talked about this a while ago. I don't remember if we had a concrete suggestion for the API. The simplest thing I could imagine is `gamepad.mapping = "";` However, then we have to decide what happens to the Gamepad object that you've just set the mapping on. Does it need to be immediately remapped? If the spec describes Gamepad objects as snapshots (https://github.com/w3c/gamepad/issues/8) does that make sense?

One quirk here is that there may be platform APIs for which this doesn't make sense, like XInput, where you don't actually get a bag of buttons + axes, the API provides you with a set of named controls. In that situation the standard mapping makes sense but unmapped doesn't really make sense.

pyalot:
There's two kinds of mappings. One is the mapping defined by the Gamepads API. There's only one mapping currently (gamepad-likes). The problems with that is:
1. UAs use guessing to figure out when to apply the mapping. This can come out wrong, for instance (real world example), I had a steering wheel mapped as a gamepad.
2. If something does get mapped as a gamepad-like, other controls (which aren't conforming to that mapping) are hidden. This is an issue for things that get (correctly) identified as gamepads, but offer additional buttons/axes that aren't fitting in the schema (for instance the PS4s touchpad)

The bottom line is that UAs can make an educated guess. But by no means is their idea always the right one.

The other kind of mapping is when the OS API provides a list of axes and buttons, which might have some kind of mapping applied. In this context (for every API, Xinput, etc.) you always get a bag of controls, but there's additional properties attached (like if it's an axis or a button and a name).

For this kind of mapping much of the same caveats apply. It's a frequent occurence to get confusing names, axes reported as buttons, buttons reported as axes, inconsistent (between OSes) descriptors, etc. Regardless of the attached metadata, what is mostly consistent on these is that if one device is plugged into one OS, the bags will be the same. (I should mention that some OSes like Linux are particularly messy there because they can talk to devices via evdev, xinput, libusb etc. and depending trough which API you go, you get completely different bags.

There's relatively little a browser can do about the underlying API other than when no mapping is possible and/or desired, to return the bags as obtained from the OS.

luser:
For XInput specifically this is what you get:
[...]

PR #25, opened 2016-06-15
Added GamepadPose

toji:
This is my proposal for adding motion, orientation, and position tracking to the Gamepad spec. Seems like creating a pull request here may be a more productive way to fostering conversation around the feature than the mailing list, which was relatively quiet on the subject.

The new interfaces are closely modeled after the VRDisplayPose and VRDisplayCapabilities interfaces defined in the [WebVR spec](https://w3c.github.io/webvr/), and I expect that they will commonly be used in conjunction. I've already prototyped exposing the device pose like this in my [experimental WebVR builds](https://webvr.info/get-chrome) and the community has taken to it fairly enthusiastically:

http://elevr.com/drawing-in-webvr/
https://twitter.com/mrdoob/status/742582533561290752
https://twitter.com/fernandojsg/status/740685191581175810
https://twitter.com/fernandojsg/status/737356455737712642
https://twitter.com/gordonnl/status/742920230498799616

Despite having it's origins in VR, however, I don't feel this features should explicitly be tied to the WebVR spec since it could benefit many more traditional gamepads as well. The Wii remote or PS4 controllers could provide accelerometer and orientation data, for example.

GamepadCapabilities is added here to make explicit wether or not a device can communicate position or orientation, since those values may not always be available continuously and it's important for developers to disambiguate between the hardware not having the capability at all (`capabilities.hasPosition = false`) or the device momentarily losing tracking because, say, a cat walked in front of an external sensor. (`pose.position = null`) There is no `hasPose` in the capabilities interface because that should be implied by the pose attribute being `null`. I would expect that this interface would be expanded in the future to cover similar capabilities (`canVibrate`, perhaps?)

Very happy to discuss this further and/or clean up the pull request to clarify things! Thanks!

### DIFF
--- a/index.html
+++ b/index.html
@@ -310,6 +310,13 @@
 
         </dd>
 
+        <dt>readonly attribute GamepadPose? pose</dt>
+        <dd>
+          The current pose of the gamepad, if suppoerted by the hardware.
+          Includes position, orientation, and acceleration. If the hardware
+          cannot supply any pose values MUST be set to null.
+        </dd>
+
       </dl>
     </section>
 
@@ -361,6 +368,85 @@
 
           For buttons without an analog sensor, only the values 0.0 and 1.0
           for fully unpressed and fully pressed MUST be provided.
+        </dd>
+      </dl>
+    </section>
+
+    <section>
+      <h2><dfn>GamepadPose</dfn> Interface</h2>
+      <p>
+        This interface defines the gamepad's position, orientation, and
+        acceleration.
+      </p>
+
+      <dl title='interface GamepadPose' class='idl'>
+        <dt>readonly attribute Float32Array? position</dt>
+        <dd>
+          Position of the gamepad as a 3D vector. Position is given in meters
+          from an origin point, which is determined by the gamepad hardware, and
+          may be the position of the gamepad when first polled if no other
+          reference point is available. The coordinate system uses these axis
+          definitions, assuming the user is holding the gamepad in the forward
+          orientation:
+
+          <ul>
+            <li>Positive X is to the user's right.</li>
+            <li>Positive Y is up.</li>
+            <li>Positive Z is behind the user.</li>
+          </ul>
+
+          MUST be null if the gamepad is incapable of providing positional data.
+          When not null MUST be a three-element array.
+        </dd>
+
+        <dt>readonly attribute Float32Array? linearVelocity</dt>
+        <dd>
+          Linear velocity of the gamepad in meters per second. MUST be null if
+          the sensor is incapable of providing linear velocity. When not null
+          MUST be a three-element array.
+        </dd>
+
+        <dt>readonly attribute Float32Array? linearAcceleration</dt>
+        <dd>
+          Linear acceleration of the gamepad in meters per second. MUST be null
+          if the sensor is incapable of providing linear acceleration. When not
+          null MUST be a three-element array.
+        </dd>
+
+        <dt>readonly attribute Float32Array? orientation</dt>
+        <dd>
+          Orientation of the gamepad as a quaternion. An orientation of
+          [0, 0, 0, 1] is considered to be "forward". The forward direction is
+          determined by the gamepad hardware, and may be the orientation of the
+          hardware when it was first polled if no other reference orientation is
+          available. MUST be null if the sensor is incapable of providing
+          orientation data. When not null MUST be a four-element array.
+        </dd>
+
+        <dt>readonly attribute Float32Array? angularVelocity</dt>
+        <dd>
+          Angular velocity of the gamepad in meters per second. MUST be null if
+          the sensor is incapable of providing angular velocity. When not null
+          MUST be a three-element array.
+        </dd>
+
+        <dt>readonly attribute Float32Array? angularAcceleration</dt>
+        <dd>
+          Angular acceleration of the gamepad in meters per second. MUST be null
+          if the sensor is incapable of providing angular acceleration. When not
+          null MUST be a three-element array.
+        </dd>
+
+        <dt>readonly attribute boolean hasPosition</dt>
+        <dd>
+          The hasPosition attribute MUST return whether the gamepad is capable
+          of tracking its position.
+        </dd>
+
+        <dt>readonly attribute boolean hasOrientation</dt>
+        <dd>
+          The hasOrientation attribute MUST return whether the gamepad is
+          capable of tracking its orientation.
         </dd>
       </dl>
     </section>


toji:
Ping! I would like to keep the conversation around this feature moving.

The device pose, whether it's a full 6DoF position and orientation in space or nothing but acceleration, is something that has been requested in the Gamepad API for a while and has proven to be a necessity for interacting with VR systems. I'm happy to discuss desired changes or alternate ways of approaching the problem, but I feel it's important that we define how to handle these type of devices one way or another. My proof-of concept for the API has proven fairly popular, and we are looking to ship this capability in Chrome in the near term.

cvan:
I'd prefer to keep the discussion in issues, but I suppose it's not a big deal. Ping @kearwood.

toji:
I have no strong opinion on where the discussion takes place, so long as it's primarily public. A pull request felt appropriate given that I had a suggested patch to the spec that would be the topic of discussion. If there's an advantage to moving the conversation elsewhere I'm happy to do so.

toji:
Yet Another Ping!

I know this represents a fairly large new feature for the API and that makes it a bit intimidating to adequately review. This is a very important addition for VR support, however, and fulfills longstanding feature requests from the community, so I'm gonna just keep on pinging till we can drum up the necessary discussion. :)

index.html
```diff
@@ -352,6 +365,94 @@
...
+    </section>
+
+    <section>
+      <h2><dfn>GamepadCapabilities</dfn> Interface</h2>
+      <p>
+        This interface defines the gamepad's hardware capabilities.
+      </p>
+
+      <dl title='interface GamepadCapabilities' class='idl'>
+        <dt>readonly attribute boolean hasPosition</dt>
```
cvan, line 440: I'm comfortable with `Gamepad#pose`, but this one not so much. can you explain other instances that are not specific to VR when we need capabilities info. that cannot be discerned from the presence of buttons and axes and their values?

toji, line 440: I felt this was necessary for the same reason we have the capabilities structure in WebVR: a null pose.position may not actually mean the device is incapable of reporting it, only that it can't report it right now. I do feel it's important to be able to differentiate those two cases.

That said, it probably is a mistake to have it in the same PR as the pose. Two proposals:
1. I move this to it's own PR so that this one can focus on pose exclusively. (Pose is still usable without the capabilities, just not as developer friendly)
2. We merge the capability booleans into the pose interface. This would mark another difference from the same interface in WebVR, but I wouldn't object to it.

mrmcpowned:
I for one am in favor of these additions. The more verbose the API can get in terms of data supplied from the gamepad, the better. Devices which would benefit from these additions would be the PS4/PS3 controllers and the Steam Controller (whenever/if ever UAs become able to access data from it), as these devices have gyrocopes.

domenic:
(Editorial review note: the rendered version is broken, and only shows the text "MUSTMUST".)

I believe I've mentioned this before somewhere, but I can't find where, so let me raise it here again. I think this pattern of using Float32Arrays is very un-idiomatic and is not something we should be doing on the platform. Here are a few reasons:
- It does not fit with the purpose of the typed array specification, which is to provide facilities for manipulating blocks of packed binary data (including with multiple views). One way of seeing this is that for most binary data, accessing the underlying buffer makes sense and is useful, whereas for these properties, `pose.position.buffer` or `pose.position.byteLength` is meaningless and confusing.
- It is awkward for authors. Other places in the platform that use points or vectors use named properties, such as `x`/`y`/`z`/`w`, which is much easier to read and work with. `pose.position[2]` is very strange.
- It means that these vectors will not benefit from existing platform APIs. For example, the `DOMPoint(ReadOnly)` classes have a very useful method, `matrixTransform`, which allows you to transform them, and have a nice pretty-printing toString() method. As we add more methods to DOMPoint(ReadOnly), those will spread throughout all places that use those classes. Cutting authors off from such future improvements is hostile.
- It has incorrect mutability semantics. You can set arbitrary elements of the typed array. What does `pose.position[2] = 5` do? What does `pose.position[99] = 10` do? What does `new Float64Array(pose.position.buffer)[0] = 1e20` do? The result is mainly confusion; presumably this doesn't actually communicate with the gamepad to update your position in some way. All of these would not be possible in a well-designed class. Using `DOMPointReadOnly` would give you the correct semantics of only reflecting the actual position (etc.), as properties with only getters.

I hope this is convincing.

toji:
Thanks for the feedback, Domenic!

In a previous version of WebVR (which this proposal is based on) we did use DOMPoints, but when we refactored we switched to using TypedArrays because it wasn't clear that the Geometry spec was actually making progress. There seemed to be a lot of debate in the mailing lists about the actual benefit of the matrix type specifically and there didn't appear to be active development or spec work happening. As such we were loath to gate our spec behind one with an apparently uncertain future. In fact, looking at https://www.chromestatus.com/features/6015941766807552 it seems that the situation is still the same on the Chrome side of things. It's listed as "No Active Development" and just today the shipping bug has someone asking about release timelines after half a year of inactivity.

If you have further insight into the status of the Geometry interfaces I'd love to hear!

Beyond that I disagree with some of the other points. The TypedArrays don't show up as cleanly in an inspector window, but there [are libraries](http://glmatrix.net) that make using them as easy as any other explicit vector type. Possibly faster, too, as one of the criticisms leveled at the Geometry interfaces was that the marshaling required to get the values into native code would negate the benefits of doing the operations natively.

It's also unlikely that most code bases will use the type provided directly no matter what we provide, as almost all 3D or other game libraries have their own concept of vector that this would need to be converted in and out of. In Three.js, for example, can convert from either an object with x/y/z/w members or an array with equal ease.

I'm also a bit concerned by DOMPoint not distinguishing between 3 and 4 element vectors, and by representing a Quaternion with a type named "Point". In either case the user will have to refer to the spec to understand the meaning of the attributes, but a generic array seems less potentially misleading.

As for the mutability argument, I do admit it's a concern. I've heard it suggested that the array properties should be marked up as [Frozen], which seems reasonable to me, but I sadly my WebIDL knowledge fails me here and I'm not sure how to properly express that in the spec. Pointers welcome!

I'm not completely against using a more explicit type for these values, but hopefully I've adequately explained why this proposal does not rely on DOMPoint. I'm happy to discuss further, as I'm sure I'm not as intimately familiar with the state of that API as some.

index.html
cvan, line 440: I'd opt for #1. In a follow-up PR, instead of a `GamepadCapabilities` that seems specific to VR, can we rename the properties or have a nested object for VR-specific properties?

What about just having `GamepadPose#position` return `undefined` if the gamepad does not support positional tracking? Do you think we can perhaps opt for returning `undefined`, `null`, and `Float32Array`s instead of introducing a new interface? (Or even maybe instead of having a separate `GamepadCapabilities` interface, could we just make `hasPosition`, `hasOrientation`, etc. be part of `GamepadPose`?

What are your thoughts?

(P.S. Sorry for the delays. Just got back from PTO.)

cvan, line 440: With that being said, feel free to merge the `GamepadPose` changes now. And let me know what you think about my suggestions re: `hasPosition` and `hasOrientation`.

toji, line 440: I'll merge those properties into the `GamepadPose`, which allows everything VR-centric to stay in the same object.

Thanks for the vote of confidence, but I can't merge on this repo. It would have to be @luser, @sgraham or another member with write access.

toji:
Okay, updated the PR to have the capability bools as part of the pose rather than a separate interface.

domenic:
The Geometry spec seems fine to me. I'm not sure why Chrome has not implemented it, but perhaps because there are only a few APIs on the platform that use it (and those are using the legacy non-standard `ClientRect` type in Chrome). I'd encourage you to implement them in Chrome as part of Web VR and the Gamepad API!

It's true that there are libraries to work around the awkwardness you are imposing on developers. That isn't how we should design the platform. Similarly, we shouldn't assume that the platform can be as user-unfriendly and cryptic as it wants because people will just wrap it in libraries.

I don't really have an opinion on representing a quaternion with a type named point. That seems like a reasonable concern, but I don't think it justifies switching to a type named array.

It's not possible to avoid the mutability while using typed arrays. There are two possibilities: fully mutable, or fully immutable (the latter using frozen). What DOMPointReadOnly gives you is readonly: i.e., mutable by the creator (the user agent), but immutable from the program. That is what is necessary for this API.

toji:
> The Geometry spec seems fine to me. I'm not sure why Chrome has not implemented it

Because, as I mentioned before, there was significant doubts about performance which have yet to be addressed. And it's not just Chrome. The thread that I found for WebKit raised the same issues, and I don't think Edge has shown any interest in the API at all. Firefox seems to be the only platform that has committed to the API.

There were also concerns about whether or not the objects presented in the spec would actually provide enough functionality to be useful. For example: The spec describes no functions for the create of a projection matrix, which is extremely frequently used by 3D graphics applications. The DOM just happens to handle that for you, so it was left out. What this means is that whether you use the Geometry spec or not users are still being forced to pull in a third party library to handle the average 3D application's needs.

> I'd encourage you to implement them in Chrome as part of Web VR and the Gamepad API!

Forgive me, but I have plenty of work to do and deadlines in which to do them. It's not practical for me to try and sufficiently address the issues in and fully ship an unrelated API in Chrome just to avoid using a structure that developers have already proven happy to work with. If I saw active forward momentum in the Geometry interfaces I may feel differently, but it seems that they have stalled out due to lack of interest.

> It's true that there are libraries to work around the awkwardness you are imposing on developers.

I expect that these values will frequently be fed into WebGL for visualization, which accepts TypedArrays as it's representation of vectors and matrices. Is working with the native type accepted by the target API really more awkward than dealing with conversions in and out of a type meant for DOM manipulation?

> It's not possible to avoid the mutability while using typed arrays.

This is a fair point, but I'm not convinced it outweighs the above concerns about the Geometry interfaces. 

At their core `DOMPoint` and `DOMMatrix` appear to be designed with the needs of CSS transforms in mind. That's great! But it doesn't appear to be well suited for the average WebGL application, which I expect to be a far more common use of this API. (And for the times when users do want to use the Geometry interfaces instead it's very easy to create one from a TypedArray with the `fromFloat32Array` constructors.)

[**EDIT**] I should also point out that we definitely don't want to preclude either this API or the WebVR API from being used with non-browser ecosystems like Node. It's not clear to me that the Geometry interfaces are compatible with that goal, given their ties to the DOM. I'd be happy to get some clarity in this area.

domenic:
Geometry APIs are fully implementable in Node. They are also somewhat implemented in other browsers under other legacy names (ClientRect etc.).

I guess we've both said our piece and from here the only thing to do is wait for a TAG (and Blink API owners) review. I still think using binary data types to represent vectors or other cases where the buffer is not meaningful is an architectural antipattern and hope that that will be considered during such review.

kearwood:
This looks good to me.  Perhaps it may be extended in the future to expose tracking systems that return "skeletons", such as Leap Motion, Kinect, and Perception Neuron.  This could be done by adding a childPoses attribute and bone identification attributes to gamepadPose.

kearwood:
Regarding Geometry API, ideally the WebVR API and GamePad API could be consistent.  The Geometry API was simply not ready yet when WebVR was first implemented and does not yet include quaternions.

Note that many the members of poses are likely to be concatenated into a large array in game engine code to be processed en-masse in a simd-like manner and then uploaded in this form to the GPU.  Perhaps float32array (or any object that can return its members as an array) would add some convenience when implementing this pattern.

cvan:
what's the status of this? can we merge? or are we waiting for 1.1?

toji:
This particular pull request is probably on hold until the current Gamepad API has reached CR status. In the meantime we're working on documenting this functionality and related ones in the supplemental [Extensions spec](https://github.com/w3c/gamepad/pull/32). The most effective place to continue this discussion is probably against that doc.

Closing this PR to avoid forking the conversation.

Issue #24, opened 2016-06-12
Consider changing gamepad-mapping API to match SDL2's


cvan:
I understand it was likely wisest to reduce the surface area of the Gamepad API when it was initially spec'd and implemented. But, today, it's cumbersome that the Gamepad API doesn't resemble SDL2's API.

emscripten-ported Unity/Unreal games to the Web are using SDL2 for input, joysticks, gamepads - and .

https://github.com/kripken/emscripten/blob/5f09184/src/library_browser.js

luser:
I think changing the API in incompatible ways is a non-starter at this point, since we have broad adoption by browsers (Gecko, Blink, Edge).

What did you have in mind here, specifically?

cvan:
The handling of [controller](https://wiki.libsdl.org/CategoryGameController) [mappings](https://wiki.libsdl.org/SDL_GameControllerMapping) (see [this crowdsourced listing [by 9 contributors]](https://github.com/gabomdq/SDL_GameControllerDB/blob/master/gamecontrollerdb.txt)) and [haptics](https://wiki.libsdl.org/CategoryForceFeedback) is mostly what I thought we could borrow from SDL.

There are Firefox bugs on file for Gamepad [mappings](https://bugzilla.mozilla.org/show_bug.cgi?id=855364) and [rumble haptic feedback](https://bugzilla.mozilla.org/show_bug.cgi?id=680289), which have good discussions. There are also spec issues on file in this repo for both of these as well: #7 and #19.

With that being said, we can probably close this. Although it breaks compatibility and arguably complicates the API surface area, I feel it could be worth a proper investigation of SDL's Gamepad API methods and capabilities to see what the Web-based API can keep the same. I'm not completely well versed in (w)asm, but the last time I was poking around, I noticed there's a lot of logic thrown away when games are exported to JS using emscripten and using the SDL Gamepad API: https://github.com/kripken/emscripten/blob/07b87426/src/library_sdl.js

I don't know if/where there are metrics for Gamepad usage in Firefox, Chrome, Edge, and Opera. The only thing I was able to find were these two Gamepad metrics from the [Chrome Platform Status > JS Usage page](https://www.chromestatus.com/metrics/feature/popularity), both at <= 0.0001% usage:
- [`PrefixedGamepad`](https://www.chromestatus.com/metrics/feature/timeline/popularity/291)
- [`DocumentCreateEventGamepadEvent`](https://www.chromestatus.com/metrics/feature/timeline/popularity/1199)

[...]

Issue #19, opened 2016-01-26
Vibrate feature for Gamepad


goldenratio:
Similar to `navigator.vibrate()`, we need vibrate feature for Gamepad API. What do you all think? 
I should be able to vibrate gamepad, if the gamepad supports vibration (example: XBOX 360)

cvan:
There was a [thread from April 2014 on the `public-webapps` mailing list](https://lists.w3.org/Archives/Public/public-webapps/2014AprJun/0046.html) about adding something like `Gamepad.vibrate`.

Also there's a [Firefox platform bug](https://bugzilla.mozilla.org/show_bug.cgi?id=680289) on file for implementation (couldn't find a bug in the [Chromium tracker](https://crbug.com/)). See [this comment by @luser in the Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=680289#c28).

goldenratio:
ok, I created new issue in Chromium tracker https://code.google.com/p/chromium/issues/detail?id=581515

Seems like no progress from Firefox side.

cvan:
Also see this mailing list thread started by @toji: https://lists.w3.org/Archives/Public/public-webapps/2016AprJun/0052.html

We ought to probably keep the discussion here now.

luser:
As with most things, I think we're still primarily blocked on figuring out a sensible API that can expose the features of modern devices without being painful to use. We wanted to try to reuse the Web Vibration API interfaces, but they are not expressive enough for the needs of the Gamepad API. Primarily:
- Multiple vibration motors per device: most modern gamepads have two, the Xbox One controller has four, and generally the motors are of varying strengths. (Usually at least one strong and one weak vibration motor.)
- Vibration motors allow setting the vibration strength, not just the duration.

dwoldrich:
@luser Thanks for all your work on Gamepad to-date, it's so neat to see most browsers now supporting it.  

I have a concern with a Gamepad Vibration/Force Feedback API that the user's JavaScript could send too many change commands too quickly for some devices to make sense of.  If HTML5 does get Gamepad Vibration, could there be some sort of error code response to say "you're sending too much data, try again later"?  Maybe there could simply be a rule that only one command could be sent per axis per time slice?  

That might be too harsh, perhaps throttling the axis update data rate could be event-driven, but in any case, I would like an API that would perform reasonably well on most hardware and operating systems by holding the developer's hand a little and not letting him abuse the device.

toji:
[...]

Issue #7, opened 2015-04-24
 Need to spec better support for control mapping


luser:
Originally filed as https://www.w3.org/Bugs/Public/show_bug.cgi?id=21387:

 Ted Mielczarek [:ted] 2013-03-25 12:27:23 UTC

The spec text that talks about a "standard gamepad" layout is a good start, but I think we need to go a little farther. We should spec a way for implementations to provide mapping data for known controllers as well as for content to fill in mappings for devices that the implementation doesn't know about.

There are two things content will want to know:
1) The mapping of controls to the physical layout of the controller. For example, most modern gamepads have a left analog stick that's the natural control for movement, and a set of buttons that are naturally under the right thumb when holding the controller. Content should be able to locate those to map them to actions.
2) The naming of controls on the actual gamepad. This is important for providing feedback to a user, content will want to be able to say "Press A".

Scott has a JS library he wrote that provides this:
https://github.com/sgraham/gamepad.js/

I've also started sketching some thoughts in data form of what a useful presentation of controller data might look like:
https://github.com/luser/gamepad-data

[reply] [−] Comment 1 Ted Mielczarek [:ted] 2013-03-25 14:27:14 UTC

To take Scott's "standard gamepad" layout a little further, I drew a diagram:
http://people.mozilla.com/~tmielczarek/gamepad.svg

This is what the data in my gamepad-data github repo is based on.

[reply] [−] Comment 2 Ted Mielczarek [:ted] 2013-03-26 12:20:52 UTC

After thinking about this quite a bit, I think this might be too much to bite off for an initial rev of the spec. I think what I'd like to do is take the "standard gamepad" approach and just polish it a little bit:
1) Add a Gamepad.mapping field, which can be set to "standard" or something if the implementation is using the standard gamepad mapping.
2) Add a way for the implementation to provide control names.

This should be enough to be useful, and we can do something more ambitious in a later version of the spec.

[reply] [−] Comment 3 Ted Mielczarek [:ted] 2013-04-10 18:35:30 UTC

I added Gamepad.mapping to the spec:
https://dvcs.w3.org/hg/gamepad/rev/c5c7b4651d4e

I'd still like to figure out how to expose control names to content, since I think that's really useful.

cvan:
Copying and pasting my comment in https://github.com/w3c/gamepad/issues/24#issuecomment-237517234:

[...]

Issue #6, opened 2015-04-24
 Should specify how to present d-pads/triggers as buttons


luser:
Originally filed as https://www.w3.org/Bugs/Public/show_bug.cgi?id=21386:

 Ted Mielczarek [:ted] 2013-03-25 12:17:01 UTC

Currently the spec requires implementations to express the set of controls on a controller in two arrays: buttons and axes. All of the platform APIs I've used expose a set of buttons and a set of axes, so this makes sense. However, there are also frequently some controls that aren't exposed as either of those, so some internal mapping is required. For example, d-pads are frequently exposed as a "POV hat" (since the USB HID was written in the days of actual flight-stick like joysticks).

I think we need to have some spec text that speaks to how to put these controls into the list of buttons/axes so that the list of controls are consistent across browsers (at least for the same OS+driver combination).

PR #26, opened 2016-06-15
Added touched property to GamepadButton

toji:
This is my proposal for adding explicit touchpad support to the Gamepad spec. Seems like creating a pull request here may be a more productive way to fostering conversation around the feature than the mailing list, which was relatively quiet on the subject.

It seems to me that effective touchpad support should only require this single new attribute. Touchpad position can accurately be represented by the existing `axes` array, but developers will need to be able to disambiguate between `[0, 0]` being a centered touch or no touch at all. That can be handled via the existing `pressed` value in a button, but it's awkward both in name and use. Many touchpads are clickable, meaning that the single control may have both a touched an a pressed state. Having a touched and a pressed state on a single button allows us to track both states while logically associating them with a single physical control. Furthermore some touchpads are pressure sensitive, but it's hard to imagine that such a control would also have separate tracking for clicked pressure, so the single existing `value` attribute should cover both touch pressure and button pressed amount.

There is also some upcoming hardware ([Oculus Touch](https://www.oculus.com/en-us/touch/)) for which many of the physical buttons will also have capacitive touch tracking to determine the position of the fingers. Adding a `touched` property to the buttons serves that case well.

There still exists some ambiguity over which button is associated with which axes, but that's a pre-existing problem and this addition doesn't make it any worse. This also ignores multi-touch surfaces (The PS4 controller has one) but considering the large amount of effort that has gone into supporting multi-touch for mobile devices it seems like that's not something we want to duplicate into the gamepad API.

### DIFF
--- a/index.html
+++ b/index.html
@@ -338,10 +338,16 @@
 
         <dt>readonly attribute boolean touched</dt>
         <dd>
-          The touched state of the button. This property MUST be true if the
-          button is capable of detecting touch and is currently touched. If the
-          button is not capable of detecting touch or is not currently touched
-          this property MUST be false.
+          The touched state of the button. If the button is capable of detecting
+          touch this property MUST be true if the button is currently being
+          touched and false otherwise.
+
+          If the button is not capable of detecting touch and is capable of
+          reporting an analog value this property MUST be true if
+          the value property is greater than zero and false if the value is 0.
+
+          If the button is not capable of detecting touch and can only report a
+          digital value this property MUST mirror the pressed value.
         </dd>
 
         <dt>readonly attribute double value</dt>
--- a/index.html
+++ b/index.html
@@ -336,6 +336,14 @@
           reasonable value.
         </dd>
 
+        <dt>readonly attribute boolean touched</dt>
+        <dd>
+          The touched state of the button. This property MUST be true if the
+          button is capable of detecting touch and is currently touched. If the
+          button is not capable of detecting touch or is not currently touched
+          this property MUST be false.
+        </dd>
+
         <dt>readonly attribute double value</dt>
         <dd>
           For buttons that have an analog sensor, this property


cvan:
> There still exists some ambiguity over which button is associated with which axes, but that's a pre-existing problem and this addition doesn't make it any worse. This also ignores multi-touch surfaces (The PS4 controller has one) but considering the large amount of effort that has gone into supporting multi-touch for mobile devices it seems like that's not something we want to duplicate into the gamepad API.

Let me preface this by saying I'm extremely pleased with the tremendous work you've done and progress you've made with your Chromium builds. And I'm glad we have browser builds today to prototype HTC Vive motion controller support with WebVR.

That aside, perhaps it's premature to introduce `GamepadButton#touched` without consideration of multitouch.

(And, apologies for the radio silence on the [mailing list thread](https://lists.w3.org/Archives/Public/public-webapps/2016AprJun/0052.html); I had a lengthy draft I never got around to sending. So thanks for bringing it up again.)

We want to know the pressure in addition to the location and distance moved of the X and Y axes. As @patrickhlauke [brought up in the mailing list thread](https://lists.w3.org/Archives/Public/public-webapps/2016AprJun/0057.html), [Pointer Events](https://www.w3.org/TR/pointerevents/) are likely a good candidate for representing the touch data from controllers.

IMO, instead of overloading `GamepadButton`, perhaps we should introduce a `GamepadAxis` interface (per my comment [here](https://lists.w3.org/Archives/Public/public-webapps/2016AprJun/0056.html)).

But this begs the question: should we continue designing an API for accessing gamepad data on the Web that diverges from native controller APIs (that I'm aware of at least)?

In particular, the [SDL2 GameController API](https://wiki.libsdl.org/CategoryGameController) already supports multitouch events, [haptic feedback](https://wiki.libsdl.org/CategoryForceFeedback) (with support for controllers with multiple haptic motors; this goes beyond the `Gamepad#vibrate(pattern)` in the current Chromium WebVR builds), and [controller mappings](https://wiki.libsdl.org/SDL_GameControllerMapping). These are all items missing from the Gamepad API spec but supported well by SDL2 APIs.

There's been a [Bugzilla bug on file for this](https://bugzilla.mozilla.org/show_bug.cgi?id=855364) for some time now. And I filed an issue about SDL2 (#24), wherein we can continue the discussion. And many of the associated issues about the current API's limitations are already filed as well: #19, #24, #7, #6, #4.

Anyway, I wanted to bring this up before we introduce too many new properties or overload the existing interfaces. Thanks!

toji:
Thanks for the comments @cvan! A couple of things I wanted to point out in response:

> We want to know the pressure in addition to the location and distance moved of the X and Y axes.

I feel like the button `value` handles this appropriately right now. I can't think of any situation in which you'd have a separate touch pressure and pressed value on the same input.

> Pointer Events are likely a good candidate for representing the touch data from controllers.

I agree! Especially for multitouch inputs, it feels like we'd need a much deeper level of API support than we likely want to add to the Gamepad API. I'm hesitant to say that _all_ touch inputs need to use a separate API like that, though. It doesn't feel appropriate to route the Vive's touchpad through PointerEvents, for example. All we really need in that case is a way to disambiguate the [0, 0] case and the rest works as-is. 

> IMO, instead of overloading GamepadButton, perhaps we should introduce a GamepadAxis interface (per my comment here).

I assume that would be used in place of floats in the `axes` array? If so I'm really reluctant to do so, since that would be a significant breaking change. Adding `touched` to button doesn't require any existing code to change and can safely be ignored by devs who don't care about the property. I'l also point out that while the Steam/Vive controllers have a convention of associating vibration with axes (because they actually vibrate the touch pads) that doesn't apply to all devices. I don't think it's a good idea to associate axis data with vibration.

I'll also point out that I haven't submitted a pull request for any haptics support, because it's the part of the experimental builds that I'm least happy with. After toying with my original API I definitely think we can do better than a simple pulse pattern for the entire gamepad, and would love to discuss that side of things further! Probably a better conversation for an issue like #19 though.

luser:
I don't think I know enough about the Vive/Steam controller hardware to understand what's going on here. Can you give me a high-level summary of what they do? I'm a bit loathe to overload bits of the existing API to support very different controllers. I'd also like to make sure we're not adding things to the API that are too specific to a single controller.

luser:
Specifically, it feels like trying to shoehorn a touchpad into the existing buttons/axes is just not a great model. Building something off of Pointer Events seems like it would be much better, since that spec has spent a lot of time describing useful semantics.

patrickhlauke:
is it naive of me to suggest that a gamepad with built-in touchpad could/should be effectively treated as two separate inputs? on the one hand, the strictly gamepad side (buttons, sticks with axes, etc) which can be queried using the gamepad API, and on the other a touchpad, which is handled using the pointer events API?

i think the one pain point here would be if there are multiple gamepads, as there's currently no concept of assigning anything other than a unique ID for each pointer - assuming the touchpad is multitouch, you could end up with lots of pointers, all with their own unique ID, but no overarching "this belongs to one hand/single input/controller device" (and even if there were, marrying up any unique identifier for a gamepad to the unique identifier for the pointers would probably present further challenges).

patrickhlauke:
having said all that, if it's the case of a touchpad which is essentially used exactly like a stick (so you are really only getting X and Y), i could see the argument of treating it exactly like a stick ("how" the actual hardware detects those X/Y is transparent to the developer, what would count is what it "acts" like)

toji:
I completely agree that we don't want to do something that only works for a single set of controllers! I'm also wary of tying gamepads too closely to a spec as large as pointer events, which has thus far been developed with no regard to this use case. I was my hope that this would be a good middle ground, but I'm very happy to have a constructive conversation about alternatives!

So to start off, I should point out a couple of the controllers that I'm aware of which could potentially take advantage of this feature.
- [Steam Controller](http://store.steampowered.com/app/353370/): Two touchpads, each only registers a single touch point and can register clicks like a button.
- [Vive Controllers](https://www.htcvive.com/us/product/): One touchpad per controller (you typically hold two). Each registers a single touch point and can register clicks like a button.
- [Oculus Touch](https://www.oculus.com/en-us/touch/): No touchpad, but has several capacitive controls to detect finger positions. (Joystick and two triggers can all detect when there is a finger on them.) Joystick can be clicked as well, like joysticks on XBox/PS controllers. Triggers are analog like XBox/PS triggers.
- [Daydream controller](https://developers.google.com/vr/concepts/controller-emulator): The controller for Google's VR system has a touchpad that registers a single touch point and registers clicks like a button. 
- [PS4 Controller](https://www.playstation.com/en-us/explore/accessories/dualshock-4-wireless-controller-ps4/): Has a touchpad can register two touch points and registers clicks like a button.
- [Shield Gamepad](https://shield.nvidia.com/support/shield-controller/faq/2): Has a touchpad, though I can't find out much about it. Appears to support one touch point. Clickable like a button.
- [WiiU Gamepad](https://en.wikipedia.org/wiki/Wii_U_GamePad): Touch sensitive screen, only registers a single touch point. Not clickable. (I really doubt this device is ever going to be exposed to the browser, though.)

There's also a sprinkling of long tail devices that have touchpads on them. Appears that very few if any support multi-touch.

So from my point of view it seems like most gamepads that have touch support use it for a single touch pad that functions more-or less like a joystick and is clickable. That general case already maps pretty well to the existing gamepad spec, with the exception of not having a clear notion of "touched". That's what prompted the idea of adding this single property to the button, as it felt unwieldy to link to a separate and fairly complex API to gain one additional bit of input. Especially considering that the Pointer events spec is event based, not polling, and there's no clear way to associate a given pointer with a given gamepad.

I agree that for devices with larger multitouch surfaces (especially something like the WiiU gamepad) pointer events seem more natural, but that seems to be the minority of gamepad uses.

luser:
Okay, so the Steam controller API seems to just expose it as a pair of axes:
https://github.com/ValveSoftware/source-sdk-2013/blob/0d8dceea4310fde5706b3ce1c70609d72a38efdf/mp/src/public/steam/isteamcontroller.h#L63

The Vive is used via the OpenVR SDK, I guess? That seems to treat everything as axes as well:
https://github.com/ValveSoftware/openvr/wiki/IVRSystem::GetControllerState (seems like an awkward API, TBH)

The Oculus Touch hardware seems to map the best to what you're proposing here, since it literally has bit flags for "user is touching this button" for each of the buttons:
https://developer.oculus.com/documentation/pcsdk/latest/concepts/dg-input-touch-touch/

There are a few other things there that I don't know how you'd map to the Gamepad API, like `ovrTouch_RIndexPointing`.

The Daydream Controller also seems to map the touch point to a pair of axes, and also exposes `isTouching`, which is true if the user is touching the touchpad:
https://developers.google.com/vr/android/reference/com/google/vr/sdk/controller/Controller

For the DualShock 4, I'm not sure how you'd map more than one touch to the Gamepad API.

Kinda hard to find any info on the Shield Gamepad.

I think the Wii U Gamepad is pretty far out of scope. :)

I guess the only real qualm I have here is that for everything but the Oculus Touch, the 'touched' property corresponds to the _touchpad_, and you're just sort of exposing it as an additional button, which feels weird. They do all appear to be clickable like analog sticks, so maybe that's not the worst thing. The Gamepad API doesn't have any concept of mapping a pair of axes and a button together into a single unit, which maybe it should since every dual-analog gamepad functionally has two sticks that are each composed of a pair of axes and a button for clicking the stick, but you're not really making things worse in that regard.

Leaving aside that I don't actually know what you _use_ the touched indicator for in VR games, I think this presents a little bit of a wart on the API, in that for non-VR controllers that don't have a touched indicator, the way you've proposed the spec text this is always going to be false, even if you're pressing the button in question, which seems odd. Do you think it'd be sensible to instead say something like "true if the button is capable of detecting touch and is currently touched, true if the button is pressed regardless of whether it is capable of detecting touch"? Sort of similar to how the spec states that for digital buttons, the `value` should be set to 0.0 or 1.0 for depressed or pressed.

patrickhlauke:
> with the exception of not having a clear notion of "touched". That's what prompted the idea of adding this single property to the button

what about `active`, borrowing very loosely from the idea of active pointers?

toji:
> There are a few other things there that I don't know how you'd map to the Gamepad API, like `ovrTouch_RIndexPointing`

I believe that those states are actually just convenience booleans to indicate common hand gestures based on the touched states of the controls. Thumb and middle finger down + index up == IndexPointing. That could be easily handled at the JS level and is (IMO) far too specific of a concept to try to expose through this API.

> I guess the only real qualm I have here is that for everything but the Oculus Touch, the 'touched' property corresponds to the touchpad, and you're just sort of exposing it as an additional button, which feels weird. They do all appear to be clickable like analog sticks, so maybe that's not the worst thing. The Gamepad API doesn't have any concept of mapping a pair of axes and a button together into a single unit...

You hit on most of my reasoning for associating it with the button instead of the axis, actually. In pretty much all cases that I can see today the touchable controls also function as a button, and the button is a single element for the control, whereas there's often multiple axes for the single control. I'm also extremely loath to change the axes array right now, as it would be backwards-compatibility breaking in a big way. (We already went through that once with buttons.) Admittedly its not the most intuitive layout, but it maps well to many of the underlying APIs and seems to work well for the reality of how these devices are designed today. And, crucially, existing code can simply ignore it's existence.

> Do you think it'd be sensible to instead say something like "true if the button is capable of detecting touch and is currently touched, true if the button is pressed regardless of whether it is capable of detecting touch"? Sort of similar to how the spec states that for digital buttons, the value should be set to 0.0 or 1.0 for depressed or pressed.

That sounds far more sensible than my proposal, actually! I can't think of any situation where you would have a control that is pressed but not touched. :)

> what about `active`, borrowing very loosely from the idea of active pointers?

I'm not completely opposed to it, but lacking the context provided by the pointer API I feel like it would be easily confused with the `pressed` state.

toji:
Ping! I would like to keep the conversation around this feature moving. Are there still significant objections to it or changes that should be made? And if we're not going to go this route to handle touchpads then can we decide on the mechanism that will be used?

luser:
I don't know about touchpads in general--I think your proposal to use a pair of axes for single-touch touchpads is probably okay enough, and maybe if it proves to be insufficient we could spec something more complicated based on PointerEvents in the future.

I think the `touched` property is reasonable, modulo my additional proposal in the previous comment. Without entirely wrangling it into spec-language, something like:
- If the button is capable of detecting touch, `touched` should be true if the button is currently touched and false if the button is not currently touched.
- If the button is not capable of detecting touch but has an analog value associated with it (e.g. it is a trigger), `touched` should be true if the analog value is greater than zero, and false if the analog value is zero.
- If the button is not capable of detecting touch and is a simple digital button, `touched` should mirror the value of `pressed` for the button.

should work and provide useful values for all controllers. Does that sound sensible?

toji:
Sounds very reasonable to me, and in fact I've already updated my Chrome proof-of-concept to match that behavior.

toji:
Updated the pull request to include language like what @luser suggested. Thanks!

luser:
Thanks for doing the work here! One other thing that might be nice is to write a W3C test for the new property.

cvan:
@toji do you want to open an issue about supporting multitouch or at least touch events (using Pointer Events, for example)?

toji:
Thanks for the review and merge! I'm happy to work on a test for this behavior!

@cvan: https://github.com/w3c/gamepad/issues/27

PR #45, opened 2016-10-31
Add Brandon Jones as spec editor

### LINKED ISSUE #44: Add Brandon as editor
@Toji is now co-editor for this spec, so should be listed amongst the editors.

toji:
Might as well have my name on the doc if I'm actually going to be editing this thing. :)

Fixes #44

### DIFF
--- a/index.html
+++ b/index.html
@@ -51,6 +51,9 @@
               { name: "Ted Mielczarek", url: "http://ted.mielczarek.org/",
                 company: "Mozilla", companyURL: "http://www.mozilla.org/",
                 w3cid: 49656 },
+	      { name: "Brandon Jones", url: "http://tojicode.com/",
+                company: "Google", companyURL: "http://www.google.com/",
+                w3cid: 87824 },
           ],
 
           otherLinks: [{


sgraham:
I just  got a notification from GitHub Pages that the Jekyll build failed, I guess after this change. Unfortunately, it's dumb and doesn't give an error message other than "Page build failed". I don't immediately see any (new) problem, but maybe the trailing comma was breaking a strict js parser?

Otherwise, it could be anywhere in the body too, so might be completely unrelated to this change.

toji:
Oh joy. Trailing comma was present on Ted's entry before, so I don't think that's it. More likely is a pre-existing change that didn't get processed until just now. (OR gh-pages just has a grudge against me specifically.)

sgraham:
Yeah, I noticed Ted's entry already has it, but maybe that's what's broken anyway? I'm not sure. I don't have Jekyll set up at the moment to try to get an actual useful message, sorry.

LJWatson:
@siusin is this something you can help with?

PR #68, opened 2018-02-23
spec vibrationActuator gamepad extension

nondebug:
Add a single vibrationActuator member to handle haptic effects that affect the whole gamepad.

### DIFF
--- a/extensions.html
+++ b/extensions.html
@@ -195,36 +195,74 @@
         <dt><dfn data-lt="playEffect()">playEffect</dfn></dt>
         <dd>
           <p>
-            <code>playEffect()</code> interprets the effect parameters in
-            <var>params</var> as a haptic effect with the specified
-            <var>type</var> and issues it to the gamepad. It returns a
-            Promise that resolves to <code>'complete'</code> once the effect
-            is complete.
+            The <code>playEffect(type, params)</code> method, when invoked, must return
+            a new promise <var>promise</var> and run the following steps in
+            parallel:
           </p>
           <p>
-            If the effect parameters are invalid, <code>playEffect()</code>
-            must return with a Promise resolved to
-            <code>'invalid-parameter'</code>. The validity of the effect
-            parameters is defined by the effect type.
+            1. Validate that the effect parameters in <var>params</var>
+            describe a valid effect of type <var>type</var>. The validity
+            may depend on the type of effect specified. If the effect is
+            invalid, resolve <var>promise</var> with <var>'invalid-parameter'</var>
+            and abort these steps.
           </p>
           <p>
-            If the effect type is not compatible with the actuator,
-            the Promise returned by <code>playEffect()</code> must be resolved
-            to <code>'not-supported'</code>.
+            2. If the current page is not visible, resolve <var>promise</var> with
+            <var>'preempted'</var> and abort these steps.
           </p>
           <p>
-            Repeated calls to <code>playEffect()</code> preempt previous calls.
-            If an effect is preempted, it must resolve its Promise to
-            <code>'preempted'</code> before the new effect is started.
+            3. If the vibration actuator is already playing an effect, resolve the
+            <var>promise</var> for the ongoing effect to <var>'preempted'</var>.
+          </p>
+          <p>
+            4. If the effect type is not compatible with the haptic actuator type,
+            resolve <var>promise</var> with <var>'not-supported'</var> and abort
+            these steps.
+          </p>
+          <p>
+            5. Issue the haptic effect to the actuator.
+          </p>
+          <p>
+            6. After the effect is complete, resolve <var>promise</var> to
+            <var>'complete'</var>.
           </p>
         </dd>
 
         <dt><dfn data-lt="reset()">reset</dfn></dt>
         <dd>
           <p>
-            <code>reset()</code> stops haptic feedback and preempts any active
-            effects. It returns a Promise that resolves to
-            <code>'complete'</code> once vibration is stopped.
+            The <code>reset()</code> method, when invoked, must return
+            a new promise <var>promise</var> and run the following steps in
+            parallel:
+          </p>
+          <p>
+            1. If the current page is not visible, resolve <var>promise</var> with
+            <var>'preempted'</var> and abort these steps.
+          </p>
+          <p>
+            2. If the vibration actuator is playing an effect, resolve the
+            <var>promise</var> for the ongoing effect to <var>'preempted'</var>.
+          </p>
+          <p>
+            3. Issue the reset command to the actuator to stop vibration.
+          </p>
+          <p>
+            4. Once vibration is stopped, resolve <var>promise</var> to
+            <var>'complete'</var>.
+          </p>
+        </dd>
+
+        <dt>Page visibility</dt>
+        <dd>
+          <p>
+            If page visibility is lost while an effect is playing, execute
+            these steps:
+          </p>
+          <p>
+            1. Resolve <var>promise</var> with <var>'preempted'</var>.
+          </p>
+          <p>
+            2. Issue the reset command to the actuator to stop vibration.
           </p>
         </dd>
       </dl>
@@ -274,19 +312,34 @@
       <dl data-dfn-for="GamepadHapticEffectType">
         <dt><dfn>dual-rumble</dfn></dt>
         <dd>
+        <p>
           A dual-rumble effect is a fixed-length, constant-intensity vibration
           effect intended for an actuator of type <code>'dual-rumble'</code>.
           Dual-rumble effects are defined by four parameters.
+        </p>
+        <p>
           <var>duration</var> sets the duration of the vibration effect in
-          milliseconds, default 0. <var>startDelay</var> sets the duration of
-          the delay after <code>playEffect()</code> is called until vibration
+          milliseconds, default 0. A negative <var>duration</var> is considered
+          invalid.
+        </p>
+        <p>
+          <var>startDelay</var> sets the duration of the delay after
+          <code>playEffect()</code> is called until vibration
           is started, default 0. During the delay interval, the actuator must
-          not vibrate. To prevent abuse, the user agent should consider an
+          not vibrate. A negative <var>startDelay</var> is considered invalid.
+        </p>
+        <p>
+          To prevent abuse, the user agent may consider an
           effect invalid if the total effect duration (the sum of the vibration
-          interval and delay interval) exceeds some reasonable value, say 5
-          seconds. <var>strongMagnitude</var> and <var>weakMagnitude</var> set
+          interval and delay interval) exceeds some maximum duration.
+          A maximum of 5 seconds is recommended.
+        </p>
+        <p>
+          <var>strongMagnitude</var> and <var>weakMagnitude</var> set
           the vibration intensity levels for the heavier and lighter ERM
           motors, normalized to the range <code>[0,1]</code>, default 0.
+          Values outside of this range are invalid.
+        </p>
         </dd>
       </dl>
     </section>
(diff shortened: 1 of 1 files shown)


extensions.html
```diff
@@ -174,30 +176,55 @@ <h2><dfn>GamepadHand</dfn> Enum</h2>
...
-            Repeated calls to <code>pulse()</code> override the previous
-            values.
+            If the effect parameters are invalid, <code>playEffect()</code>
+            must return with a Promise resolved to
+            <code>'invalid-parameter'</code>. The validity of the effect
+            parameters is defined by the effect type.
+          </p>
+          <p>
+            If the effect type is not compatible with the actuator,
+            <code>playEffect()</code> must immediately return a Promise resolved
```
reillyeon: The use of the word "immediately" seems to imply that if I execute the following,

```
actuator.playEffect(invalid_params).then(() => {
  console.log("1");
});
Promise.resolve().then(() => {
  console.log("2");
});
```

Then it is guaranteed to print,

```
1
2
```

In my experience APIs returning Promises run the steps that result in the resolution of the promise "in parallel" which means that the exact timing of when the promise resolves is not directly tied to Javascript execution and could, for example, first require a round-trip to between a renderer and browser process before the result is known.

extensions.html
```diff
@@ -206,28 +233,82 @@ <h2><dfn>GamepadHapticActuator</dfn> Interface</h2>
...
+          milliseconds, default 0. <var>startDelay</var> sets the duration of
+          the delay after <code>playEffect()</code> is called until vibration
+          is started, default 0. During the delay interval, the actuator must
+          not vibrate. To prevent abuse, the user agent should consider an
+          effect invalid if the total effect duration (the sum of the vibration
+          interval and delay interval) exceeds some reasonable value, say 5
+          seconds. <var>strongMagnitude</var> and <var>weakMagnitude</var> set
+          the vibration intensity levels for the heavier and lighter ERM
+          motors, normalized to the range <code>[0,1]</code>, default 0.
+        </dd>
```
reillyeon, line 331: Is it valid for any of these parameters to be 0?

nondebug: Yes, 0 is valid for all parameters. A dual-rumble effect with the default parameters describes a 0-length effect with no vibration from either motor. In practice it behaves the same as calling reset().

extensions.html
nondebug: I think this should be "in parallel". I removed the requirement that the promise is immediately resolved.

extensions.html
```diff
@@ -209,7 +209,7 @@ <h2><dfn>GamepadHapticActuator</dfn> Interface</h2>
           </p>
           <p>
             If the effect type is not compatible with the actuator,
-            <code>playEffect()</code> must immediately return a Promise resolved
+            the Promise returned by <code>playEffect()</code> must be resolved
```
reillyeon: This still somewhat implies that when playEffect() returns the promise has already been resolved. To clarify this I would start above by saying, "It returns a new Promise `promise` and performs the following steps in parallel." Then below list out the available failure cases which all "resolve `promise` with..." and the final successful case which "resolves `promise` with `'complete'`."

extensions.html
```diff
@@ -195,36 +195,74 @@ <h2><dfn>GamepadHapticActuator</dfn> Interface</h2>
         <dt><dfn data-lt="playEffect()">playEffect</dfn></dt>
         <dd>
           <p>
-            <code>playEffect()</code> interprets the effect parameters in
-            <var>params</var> as a haptic effect with the specified
-            <var>type</var> and issues it to the gamepad. It returns a
-            Promise that resolves to <code>'complete'</code> once the effect
-            is complete.
+            The <code>playEffect(type, params)</code> method, when invoked, must return
+            a new promise <var>promise</var> and run the following steps in
```
reillyeon: "new Promise"

extensions.html
```diff
@@ -195,36 +195,74 @@ <h2><dfn>GamepadHapticActuator</dfn> Interface</h2>
...
         </dd>
 
         <dt><dfn data-lt="reset()">reset</dfn></dt>
         <dd>
           <p>
-            <code>reset()</code> stops haptic feedback and preempts any active
-            effects. It returns a Promise that resolves to
-            <code>'complete'</code> once vibration is stopped.
+            The <code>reset()</code> method, when invoked, must return
+            a new promise <var>promise</var> and run the following steps in
```
reillyeon: "new Promise"

reillyeon (approved):
LGTM but capitalize type names.

extensions.html
```diff
@@ -195,36 +195,74 @@ <h2><dfn>GamepadHapticActuator</dfn> Interface</h2>
...
+            parallel:
+          </p>
+          <p>
+            1. If the current page is not visible, resolve <var>promise</var> with
```
cvan: do you want to make these paragraphs an `<ol>`?

extensions.html
```diff
@@ -274,19 +312,34 @@ <h2><dfn>GamepadHapticEffectType</dfn> Enum</h2>
       <dl data-dfn-for="GamepadHapticEffectType">
         <dt><dfn>dual-rumble</dfn></dt>
         <dd>
+        <p>
```
cvan: nit: can you indent this one by more level?

luser:
This all looks pretty sensible for handling the common case of a gamepad with weak+strong rumble motors, and it seems like it'll provide some extension room for handling more interesting things. I don't know of a lot of other haptic designs in the wild aside from:
* The Xbox One controller, which has the usual dual rumble motors + two more "impulse triggers", one near each trigger button. In the past I hadn't seen any API that actually exposed them (since I don't have access to the Xbox One SDK), but it looks like Microsoft exposes them via [the WinRT Gamepad APIs](https://docs.microsoft.com/en-us/uwp/api/windows.gaming.input.gamepadvibration) nowadays.
* The Nintendo Switch Joy-Cons, which have "HD rumble", which is apparently provided by a [linear resonant actuator](https://blog.somaticlabs.io/how-it-works-linear-resonant-actuators/) and the API involves loading waveforms representing the haptic (an LRA is functionally very similar to an audio speaker).
* The Steam Controller also has LRAs, one under each touchpad, but they seem to be dual-axis actuators. ([The iFixit teardown](https://www.ifixit.com/Teardown/Steam+Controller+Teardown/52578#s117509) says the touchpads are [Cirque GlidePoint TM040040s](http://www.cirque.com/glidepoint-circle-trackpads), and the Cirque page mentions that those are compatible with [Alps HAPTIC Reactor drivers](http://www.alps.com/prod/info/E/HTML/Haptic/), which look like what's in the photo.) I'm not familiar with the Steam Controller API, but the API contains methods like [`TriggerHapticPulse`](https://partner.steamgames.com/doc/api/ISteamController#TriggerHapticPulse), which seems to offer only a simplistic rumble-like interface to the LRAs. There's also a [`TriggerVibration`](https://partner.steamgames.com/doc/api/ISteamController#TriggerVibration) which is a very simple interface that simply lets you turn on the left and right rumble motors for a controller for a period of time.
* Old-style force feedback joysticks, which had [super complicated APIs](https://msdn.microsoft.com/en-us/library/windows/desktop/ee417563(v=vs.85).aspx) for creating and playing effects. I'm not sure if controllers like this are even manufactured anymore.

In any event, most of those seem like they could be handled with future extensions to this API. The only thing I'd worry about is that if code in the wild adds support for `dual-rumble` haptics but then the spec gets extended to add something more complex (even just the 4 rumble motors in the Xbox One, say) there doesn't seem to be a fallback path so that existing code can work with newer devices.

What if instead of having `readonly attribute GamepadHapticActuatorType type` on `GamepadHapticActuator` we instead had something like `bool canPlayEffectType(GamepadHapticEffectType type)` or a WebIDL setlike type that gave the set of `GamepadHapticActuatorType` or `GamepadHapticEffectType` values that were supported? With an interface like that, users could detect if a device supports `dual-rumble` and then use it, and if support for a more complex haptic is added to the spec and implemented by browsers, browsers could have the device support *both* the newer type *and* the `dual-rumble` type, so existing code would work with newer devices, and new code could feature-detect for `quad-rumble` or whatever if it wanted to support that.

qdot:
> Old-style force feedback joysticks, which had super complicated APIs for creating and playing effects. I'm not sure if controllers like this are even manufactured anymore.

I'm not sure if flight and wheel joysticks still use these, as most of those use USB so latency isn't as much of an issue as it used to be. However, I think we may see a return to this with HD haptics, since the timing can be precise enough that having parametric effects might be useful. I think that may be beyond the scope of "gamepads", but with the way gamepads are being repurposed for VR and including other sensors, it's kinda hard to tell?

> The Nintendo Switch Joy-Cons, which have "HD rumble"

The Nintendo Switch API seems to expose "amplitude" and "frequency" versus a simple "low/high power level" like XInput. More info can be found at:

dekuNukem/Nintendo_Switch_Reverse_Engineering#7
dekuNukem/Nintendo_Switch_Reverse_Engineering#11
https://github.com/dekuNukem/Nintendo_Switch_Reverse_Engineering/blob/master/rumble_data_table.md

And there's Haptics/Protocol code (C++) at:

https://github.com/mfosse/JoyCon-Driver/blob/master/joycon-driver/include/Joycon.hpp

I could see that getting sort of "dumbed down" the same way the Steam API does it though, which would basically wedge it into working under the proposed API.

Returning to the spec, the use of "HapticActuator" versus "VibrationActuator" is a little confusing. Depending on how much future-proofing needs to happen in the spec, is it useful to make sure these terms are divided up correctly? Vibration is only one type of actuation, but this doesn't really take other mechanisms like temperature, pneumatics, etc into account, which could show up in the future. Not sure if that really needs to be accounted for though, and definitely don't want to turn this bug into bikeshedding around that.

nondebug:
Thanks for the comments!

I agree, a canPlayEffectType method would work well and would allow the actuator type to be removed. I can definitely see the actuator type becoming an issue, especially with actuators that can reasonably be classified as multiple types. Done.

Regarding haptic vs vibration naming, I chose vibrationActuator for the Gamepad member name because it should be limited only to effects that vibrate the whole device. Rumble is common enough that I think it's worth having an explicit member to signal the capability rather than hiding it in a hapticActuators array. But, I expect we will eventually also want an array to cover other types of vibration and other types of actuators. I left it out only because it's not useful yet and would unnecessarily complicate the spec.

For instance, Xbox One trigger rumble is a vibration effect that independently vibrates one element on the gamepad. I think it would be appropriate to expose these as separate actuators in a hapticActuators array and define a canonical haptics mapping in the Standard Gamepad. (I'd prefer this over adding "quad-rumble") Similarly, if you have dual Joy-Cons or a Switch Pro controller, LRA vibration effects can be independently controlled between the two hands. I'd rather expose left and right GamepadHapticActuators that each control one Joy-Con than add a "stereo-switch-lra" effect type.

In that context, does it make more sense that the vibration-focused actuator would be vibrationActuator, and the more general collection would be hapticActuators? Calling it "vibrationActuator" has one more benefit that I find appealing, "vibration" is more obviously connected to rumble effects so I expect developers will try it first.

This API is likely not sufficient for waveform-based haptics. It should be possible to upload an effect to the device independently of playing the effect. The waveform data might be large and uploading it could incur a delay or use battery power. And because the storage of such effects isn't infinite, it should also be possible to query if an effect will fit and discard effects from the device once they are no longer needed. The use case I think of is a vibration effect based on an audio track. Each effect is a unique segment of the audio and can be discarded immediately after playing, but multiple effects can be buffered in advance for smooth playback with no gaps.

It's probably not sufficient for joysticks with directional haptics that react to the stick's position and velocity, either. I haven't thought about these too much, IMO this joystick-specific feature is not sufficiently gamepad-like to worry about here. It would be better if we could support these legacy devices through Web USB.

LJWatson:
Is this PR something to defer to v2? Ping @Sagoston and @SGraham

marcoscaceres:
Fixed merge conflict.

PR #78, opened 2018-08-01
Fix Promise type

saschanaz:
Also replaced `void` with `boolean` because [the prose says](https://w3c.github.io/gamepad/extensions.html#gamepadhapticactuator-interface):

>pulse() applies a value to the actuator for duration milliseconds. The value passed to pulse() is clamped to limits defined by the actuator type. The returned Promise will resolve true once the pulse has completed.

### DIFF
--- a/extensions.html
+++ b/extensions.html
@@ -80,7 +80,7 @@
           wg:           "Web Platform Working Group",
 
           // URI of the public WG page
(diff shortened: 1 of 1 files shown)


saschanaz:

marcoscaceres (approved):

PR #96, opened 2019-08-05
Do not permit unique info in ID

### LINKED ISSUE #73: id field in gamepad might have a persistent identifier?
(See also the email thread)

I have read in the past that the id field sometimes contains things like a serial number. Obviously this presents a very persistent tracking identifier.

Mozilla says: in Firefox it will contain three pieces of information separated by dashes (-):
Two 4-digit hexadecimal strings containing the USB vendor and product id of the controller
The name of the controller as provided by the driver. https://developer.mozilla.org/en-US/docs/Web/API/Gamepad/id

How is this exposed in other browsers? It seems like it would be advantageous to require this string to _not_ contain uniquely identifying information and to Non-normatively suggest an algorithm to do so.

Florian suggest that we standardize the id attribute so the vendor and device id are in a consistent format and order.

He also indicated that the product name is displayed to the user to help them identify the control (they won't know vendor/device id of course.)

My concern is still that the product name may contain serial numbers or something similar that would uniquely identify a particular controller vs merely identifying it as that type of controller. Have other people seen this behavior? It seems it would be near-impossible to fix on the controller side (since it'd require controllers to be reflashed with new firmware...)

sagoston:
Closes #73

Implementation commitment:

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/pull/96.html" title="Last updated on Aug 5, 2019, 8:51 PM UTC (a833e74)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/96/37d4804...a833e74.html" title="Last updated on Aug 5, 2019, 8:51 PM UTC (a833e74)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -241,7 +241,7 @@
           An identification string for the gamepad. This string identifies the
           brand or style of connected gamepad device. Typically, this will
           include the USB vendor and a product ID.
-          This string MUST NOT contain any information indentifying a unique
+          This string MUST NOT contain any information identifying a unique
           gamepad device. 
         </dd>
         <dt>
--- a/index.html
+++ b/index.html
@@ -241,6 +241,8 @@
           An identification string for the gamepad. This string identifies the
           brand or style of connected gamepad device. Typically, this will
           include the USB vendor and a product ID.
+          This string MUST NOT contain any information indentifying a unique
+          gamepad device. 
         </dd>
         <dt>
           <dfn>index</dfn> attribute


tomrittervg:
Typo: 'indentifying'

nondebug:
> An identification string for the gamepad. This string identifies the brand or style of connected gamepad device. Typically, this will include the USB vendor and a product ID. This string MUST NOT contain any information identifying a unique gamepad device. 

It's not clear to me whether existing browsers would need to change their behavior to comply with the new spec language. The original bug mentioned that some devices expose serial numbers in the product name. Would the spec change require that the product name not be included in the ID string?

sagoston:
> > An identification string for the gamepad. This string identifies the brand or style of connected gamepad device. Typically, this will include the USB vendor and a product ID. This string MUST NOT contain any information identifying a unique gamepad device.
> 
> It's not clear to me whether existing browsers would need to change their behavior to comply with the new spec language. The original bug mentioned that some devices expose serial numbers in the product name. Would the spec change require that the product name not be included in the ID string?

I don't think this is a problem.  I've not seen any browser/device combo which emits a unique serial/mac address so far.  Which device/browser has been identified to expose unique data?  Or was the speculation that it could be a problem?

nondebug:
HTML5 Gamepad Tester logs ID strings, which is helpful for collecting IDs of less common devices. Jumping to the end of the list reveals device IDs that were seen only once.

https://html5gamepad.com/controllers

The device with VID:PID 0483:572b exposes garbage data for the product name which might be a serial number:

ẋ嚕冕쁁攥犡䋀几훝쭪磌蛱홧�匟ꢦ�↝䦍ῳ (Vendor: 0483 Product: 572b)

There are also several entries with product name "Virtual Gamepad" followed by a unique string:

0001-0001-Virtual Gamepad: 3q9E8kho9

sagoston:
> HTML5 Gamepad Tester logs ID strings, which is helpful for collecting IDs of less common devices. Jumping to the end of the list reveals device IDs that were seen only once.
> 
> The device with VID:PID 0483:572b exposes garbage data for the product name which might be a serial number:
> 
> ẋ嚕冕쁁攥犡䋀几훝쭪磌蛱�홧�匟ꢦ�↝䦍ῳ (Vendor: 0483 Product: 572b)
> 
> There are also several entries with product name "Virtual Gamepad" followed by a unique string:
> 
> 0001-0001-Virtual Gamepad: 3q9E8kho9

These "Virtual Gamepads" are getting reported with "Virtual Gamepad: 3q9E8kho9" as the name of the device type/class.  All the actual gamepad controller devices report only the type, not the unique instance.  Approximately 20 uses of "Virtual Gamepad" are the exception.

As for the STMicro device (Vendor: 0483 Product: 572b), I don't know if that's garbage or something significant

tomrittervg:
Why was this closed?

sagoston:
We're making activation of the gamepad more explicit (requiring button state to go from unpressed to pressed) and adding a feature policy to only explicitly allow iframes use of the gamepad.  With these changes, requiring uniqueness in the id becomes less critical.

PR #106, opened 2019-09-19
Clarification of gamepad state from getGamepads() and connect/disconnect events

### LINKED ISSUE #83: Gamepad event timing
The spec describes gamepadconnected and gamepaddisconnected events but does not describe the timing of when these events should be dispatched or how they interact with the navigator.getGamepads polling method. I'd like to propose stricter guarantees on this timing.

To ensure there is no difference between accessing the gamepad state through event.gamepad or navigator.getGamepads() from an event listener:

> If a gamepad event is dispatched and there is an event listener for the event, the gamepad state returned by navigator.getGamepads() MUST match the gamepad state in event.gamepad. The gamepad state returned by navigator.getGamepads() MUST remain the same until all event listeners have completed.

(Possibly this should also specify that the Gamepad object is the same object as in the getGamepads array.)

To ensure a sane ordering of events so that event listeners can accurately track which gamepads are currently connected:

> gamepadconnected and gamepaddisconnected MUST NOT be dispatched twice in a row for the same gamepad. For instance, if a gamepad is determined to have been disconnected and connected in the same polling update then gamepaddisconnected MUST be dispatched first.

(Possibly also specify that gamepaddisconnected must not be dispatched before gamepadconnected for a particular gamepad, although this may already be implied by the spec.)

sagoston:
Closes #83

Implementation commitment:

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/pull/106.html" title="Last updated on Oct 10, 2019, 11:10 PM UTC (e590aaa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/106/0a78ee1...e590aaa.html" title="Last updated on Oct 10, 2019, 11:10 PM UTC (e590aaa)">Diff</a>


marcoscaceres (approved):
This seems fine, but we need to put into an algorithm. I'd be ok with landing this initially tho. 

@JamesHollyer @nondebug, how do you want to handle this?

JamesHollyer:
Yup, this is actually verbiage that I had given at TPAC.  This aligns with the algorithm that we have written up so it is in the right direction.  @nondebug and I agree that we should land this as is for now.

PR #107, opened 2019-09-19
Define user activation and added references in getGamepads and gamepa…

### LINKED ISSUE #72: Clarifications for getGamepads / user activations
a) Why, in the example, is there a leading 'null'?  Is it indicating
there are two gamepads but not giving you information about the first
one? Why?

b) "Gamepads MUST only appear in the list if they are currently
connected to the user agent, and at least one device has been
interacted with by the user." - that's great. But what does
"interacted with by the user" mean? Ever since process start? For this
origin?

knyg:
…dConnected.

Closes #72 

Implementation commitment:

***
<a href="https://pr-preview.s3.amazonaws.com/knyg/gamepad/pull/107.html" title="Last updated on May 22, 2020, 4:38 AM UTC (8c12356)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/107/33e64a4...knyg:8c12356.html" title="Last updated on May 22, 2020, 4:38 AM UTC (8c12356)">Diff</a>


sagoston (approved):

marcoscaceres (changes requested):
This is capturing the right idea, but I think we probably want to layer this onto of HTML's "triggered by user activation". I'll propose some alternative wording.

marcoscaceres:
Note to self: add "triggered by gamepad activation".

marcoscaceres:
This is generally a good idea. Talking with @nondebug and other folks on the call, it would be good if it generated a true activation event as part of HTML... that could be used for going fullscreen.

marcoscaceres:
(I suggest we close this pull request and send something new that covers the use cases that we are interested in)

PR #109, opened 2019-09-20
Remove Travis, use GitHub Action

sidvishnoi:
Might take a few runs to make it correct.


marcoscaceres:

.github/workflows/pr-push.yml
```diff
@@ -0,0 +1,32 @@
...
+
+jobs:
+  ci:
+    name: Validate and Publish
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 10
```
marcoscaceres: There is no LTS still, right?

marcoscaceres:
@sidvishnoi, where do we actually set the secrets? I couldn't find it in the settings or the actions page?

.github/workflows/pr-push.yml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Run respec-validator
```
marcoscaceres:
```suggestion
      - name: Do spec validation
```

.github/workflows/pr-push.yml
```diff
@@ -0,0 +1,32 @@
...
+        run: npx respec-validator --gh-user=$GH_USER --gh-token=$GH_TOKEN index.html
+        env:
+          GH_USER: ${{ secrets.GH_USER }}
+          GH_TOKEN: ${{ secrets.GH_TOKEN }}
+      - name: Publish
```
marcoscaceres:
```suggestion
      - name: Publish to /TR/
```

.github/workflows/pr-push.yml
sidvishnoi: Nope. Only numbers are supported presently

sidvishnoi:
Missed review notification from this.

> where do we actually set the secrets?

https://github.com/w3c/gamepad/settings/secrets should work, given you've right permissions.

.github/workflows/pr-push.yml
sidvishnoi: Maybe we should use v12 now, as it'll be LTS within a month.

marcoscaceres: sounds like a plan!

sidvishnoi:
```suggestion
          node-version: 12
```

marcoscaceres (changes requested):
Ok, all the tokens are in place... one nit... could you put the `env` before the `run`... that way, it's more clear where things are defined and the order is more logical and matches actual execution order: the "environment is set, then things are run" and so on...

sidvishnoi:
Close in favour of https://github.com/w3c/gamepad/pull/115

PR #112, opened 2019-09-23
Add permission policy integration

### LINKED ISSUE #119: Feature policy integration
In #73, @marcoscaceres wrote:

> So what if we added a feature policy to prevent third parties from accessing game pads unless the top-level document explicitly says it's ok? That would mitigate some of the privacy concerns and limit some of the damage.

There was consensus that it would be helpful.

marcoscaceres:
Closes #119 

Related: https://github.com/w3c/permissions/pull/200

 * [x] [Modified Web platform tests](https://github.com/web-platform-tests/wpt/commit/7075d7ebe1cb282c992635defc0dbdadb69c99df#diff-6a16637710567beeef8c33136cb3f28920f90afe30c518ccab78b59a908e4ce1)

Implementation commitment:

 * [ ] WebKit (https://bugs.webkit.org/show_bug.cgi?id=230136)
 * [ ] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=1011006)
 * [ ] [Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=1640086)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/pull/112.html" title="Last updated on Oct 29, 2020, 10:57 PM UTC (e57327b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/112/451286d...e57327b.html" title="Last updated on Oct 29, 2020, 10:57 PM UTC (e57327b)">Diff</a>


JamesHollyer (changes requested):
Gamepad events also need protection.  We should simply put something like "Events will not be fired unless the feature policy "gamepad" is enabled."

marcoscaceres:
> Gamepad events also need protection. We should simply put something like "Events will not be fired unless the feature policy "gamepad" is enabled."

good point!  I forgot they fire on the Window object.

JamesHollyer:
I would like to see an issue number that gets closed by this change.  But the change itself looks good. 
 Also fyi, I am currently implementing the metric for gamepad use in cross origin sub frames so we have more info on how to move forward with this breaking change.

JamesHollyer (approved):
I would like to see an issue number that gets closed by this change. But the change itself looks good to me.
Also fyi, I am currently implementing the metric for gamepad use in cross origin sub frames so we have more info on how to move forward with this breaking change.

marcoscaceres:
> I would like to see an issue number that gets closed by this change. 

#119 - once we have the telemetry data, we can move on the tests and implementation commitments.

index.html
```diff
@@ -789,6 +795,25 @@ <h3>
...
+      <h2>
+        Integration with Feature policy
+      </h2>
+      <p>
+        This specification defines a policy-controlled feature identified by
+        the string "`gamepad`". Its <a>default allowlist</a> is '`self`'.
+      </p>
+      <div class="note">
+        <p>
+          A <a>document</a>’s <a>feature policy</a> determines whether any
```
sidvishnoi:
```suggestion
          A <a>document</a>’s [=Document/feature policy=] determines whether any
```

greggman:
broke a few hundred or thousand games on itch.io

nondebug:
Which browser? As far as I know this change isn't enforced yet in Chrome or Firefox.

Can you link to an example of a game that was broken?

greggman:
You're right. There's just a warning. So they'll all break in a few months but not today

reillyeon:
According to [their documentation](https://itch.io/docs/creators/html5) itch.io hosts web games by embedding an iframe. This is why this change will break a number of games but also means that the fix is simple. itch.io needs to add "gamepad" to the existing `allow` attribute in the `<iframe>` tags on their pages. They already appear to allow a number of other capabilities through this attribute so it seems that they understand how Permissions Policy works. If I'm correct then this shouldn't require any changes on the part of individual game authors. The change can be made unilaterally by itch.io.

subsoap:
I have contacted the itch.io admin about this via e-mail. If the change is not done in a few days someone else should try pinging him via social media until he notices. He has been responsive about adding needed permissions before so it will likely be done before this is enforced.

greggman:
Anyone updating mdn?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Feature_Policy

reillyeon:
Pinging @jpmedley for the MDN update. This change hasn't shipped in Chrome or Firefox but we've started to show a console warning so a note in MDN would be helpful as well.

subsoap:
Gamepad permission has now been added to itch.io iframe. ✨

byersdz:
I've added an allow="gamepad" attribute to the iframes on https://bitmelo.com however I am still getting this warning in Chrome. Am I handling this correctly or is there something else I need to do?

nondebug:
@byersdz I tried [Katana](https://bitmelo.com/user/david/games/katana) in Chrome 88.0.4324.182 (Official Build) (64-bit) but didn't see any gamepad-related console messages. I checked the source and everything seems fine.

Can you share the text of the console warning?

reillyeon:
I saw the message when I loaded https://bitmelo.com, I think I understand the issue and have filed https://crbug.com/1183491 to track the fix.

subsoap:
Is this required yet? When adding the permission to iframes (allow="gamepad") it still prints 

> Unrecognized feature: 'gamepad'.

But it also still mentions

> VM17:1 getGamepad will now require a Permission Policy. Please update your application accordingly. For more information see https://github.com/w3c/gamepad/pull/112

Seen here in console

https://www.pixcade.com/big-klondike/

marcoscaceres:
@subsoap, I believe it should be "coming soon". @JamesHollyer, any update?

marcoscaceres:
Filed WebKit bug https://bugs.webkit.org/show_bug.cgi?id=230136 ... will implement that.

PR #120, opened 2019-10-19
Require secure context for gamepad state and events

### LINKED ISSUE #113: Restrict getGamepads() to [SecureContext]
We should evaluate that impact of making `getGamepads()` only available in `SecureContext` (as we did with Geolocation). 

We might need to figure out some kind of deprecation timeline for the `http`/insecure API.

nondebug:
Closes #113

 * [x] Modified Web platform tests - will be updated automatically after merge. 

Implementation commitment:

 * [x] WebKit (https://bugs.webkit.org/show_bug.cgi?id=215020)
 * [x] [Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=1591329)

***
<a href="https://pr-preview.s3.amazonaws.com/nondebug/gamepad/pull/120.html" title="Last updated on Oct 29, 2020, 10:59 PM UTC (a67076f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/120/792597d...nondebug:a67076f.html" title="Last updated on Oct 29, 2020, 10:59 PM UTC (a67076f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -121,7 +121,7 @@
       }
     </style>
   </head>
-  <body>
+  <body data-cite="secure-contexts">
     <section id='abstract'>
       <p>
         The Gamepad specification defines a low-level interface that represents
@@ -212,7 +212,7 @@
         This interface defines an individual gamepad device.
       </p>
       <pre class="idl" data-cite="HR-TIME">
-        [Exposed=Window]
+        [Exposed=Window, SecureContext]
         interface Gamepad {
           readonly attribute DOMString id;
           readonly attribute long index;
@@ -321,7 +321,7 @@
         device.
       </p>
       <pre class="idl">
-        [Exposed=Window]
+        [Exposed=Window, SecureContext]
         interface GamepadButton {
           readonly attribute boolean pressed;
           readonly attribute boolean touched;
@@ -428,6 +428,12 @@
           each Gamepad present at the index in the array specified by its
           {{Gamepad/index}} attribute. Array indices for which there is no
           connected Gamepad with the corresponding index should return null.
+
+          <p>
+          If the <a>environment settings object</a> is a <a>non-secure
+          context</a> return an empty array. Optionally, inform the developer
+          that the API can only be used from <a>secure contexts</a>.
+          </p>
           
           <p>
           The gamepad state returned from getGamepads() does not reflect
@@ -455,7 +461,7 @@
         <dfn>GamepadEvent</dfn> Interface
       </h2>
       <pre class="idl" data-cite="DOM">
-        [Exposed=Window]
+        [Exposed=Window, SecureContext]
 
         interface GamepadEvent: Event {
           constructor(DOMString type, GamepadEventInit eventInitDict);
@@ -757,6 +763,10 @@
         page was loaded, the <a>gamepadconnected</a> event SHOULD be dispatched
         when the user presses a button or moves an axis.
       </p>
+      <p>
+        A <a>user agent</a> MUST NOT dispatch this event type if the
+        <a>environment settings object</a> is a <a>non-secure context</a>.
+      </p>
     </section>
     <section>
       <h3 id="event-gamepaddisconnected">
@@ -775,6 +785,10 @@
         <a>user agent</a> has previously dispatched a <a>gamepadconnected</a>
         event for that gamepad to a window, a <a>gamepaddisconnected</a> event
         MUST be dispatched to that same window.
+      </p>
+      <p>
+        A <a>user agent</a> MUST NOT dispatch this event type if the
+        <a>environment settings object</a> is a <a>non-secure context</a>.
       </p>
     </section>
     <section>
--- a/index.html
+++ b/index.html
@@ -413,34 +413,35 @@
       </pre>
       <dl>
         <dt>
-          <dfn>getGamepads</dfn>
+          <dfn>getGamepads()</dfn>
         </dt>
         <dd>
-          Retrieve a snapshot of the data for the the currently connected and
-          interacted-with gamepads. Gamepads MUST only appear in the list if
-          they are currently connected to the <a>user agent</a>, and at least
-          one device has been interacted with by the user. If no devices have
-          been interacted with, devices MUST NOT appear in the list to avoid a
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -413,28 +413,35 @@ <h2>
...
-          malicious page from fingerprinting the user. The length of the array
-          returned MUST be one more than the maximum index value of the Gamepad
-          objects returned in the array. The entries in the array MUST be the
-          set of Gamepad objects that are visible to the current page, with
-          each Gamepad present at the index in the array specified by its
-          {{Gamepad/index}} attribute. Array indices for which there is no
-          connected Gamepad with the corresponding index should return null.
-          
+          <p class="note">
+            The gamepad state returned from {{Navigator/getGamepads()}} does
```
marcoscaceres, line 400: This will be nice when we actually get the gamepad manager is place - so it can just be a note IMO.

index.html
```diff
@@ -413,28 +413,35 @@ <h2>
...
+            not reflect disconnection or connection until after the
+            <a>gamepaddisconnected</a> or <a>gamepadconnected</a> events have
+            fired.
+          </p>
           <p>
-          The gamepad state returned from getGamepads() does not reflect
-          disconnection or connection until after the <a>gamepaddisconnected</a>
-          or <a>gamepadconnected</a> events have fired.
+            If the <a>environment settings object</a> is a <a>non-secure
```
marcoscaceres: Moved this so it's the first thing we check. This should be easy to test too.

marcoscaceres:
Checked the cross references locally also... they are all good 👍

marcoscaceres (approved):
Shifted order of things... let me know what you think!

marcoscaceres:
Gecko tracking bug https://bugzilla.mozilla.org/show_bug.cgi?id=1591329

marcoscaceres:
About the "secure context" not being linked on TR, filed a bug:
https://github.com/w3c/webappsec-secure-contexts/issues/68

index.html
nondebug: Makes sense, lgtm

index.html
nondebug: lgtm

marcoscaceres:
Quick updated: landed the console message + returning the empty array + one fire events in secure context in Firefox Nightly. I still need to do the events, but they were breaking a bunch of WebXR things in Gecko, so will need to come back to those.

beidson:
Adding the logging in WebKit here:
https://bugs.webkit.org/show_bug.cgi?id=214995

Restricting to SecureContext here:
https://bugs.webkit.org/show_bug.cgi?id=215020

andrewfstratton:
Please can you still allow API to work for local development (and learning)?!

marcoscaceres:
Hi @andrewfstratton. In most (all?) browsers, localhost is considered a secure context for development purposes:

https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts#when_is_a_context_considered_secure

Miniontoby:
[moved to the original issue]

BlobTheKat:
What is the rationale behind this decision?

marcoscaceres:
Tracking scripts poking at gamepads, plus also preventing random third-parties / proxies injecting scripts over HTTP that try to poke/read at gamepads.

Miniontoby:
Personally it would be nice if the whole localnet could be whitelisted for these secure context limitations, so not only localhost.

With localnet I mean your intranet, like every device connected to your router. Like every `192.168.***.***` ip or `*.fritz.box` or whatever local dns domain you have.

Since within localnet you cannot have ssl certs that are trusted. And not everyone can trust self signed certs, like at my home, meaning that they will never be able to play using gamepads on locally made web games.

*And yes, I know this might not be the best place to ask this, but because right now it is for me mostly about the gamepads, I just asked it here, but it would be nice if every feature would be possible within localnet (e.g. GPS, Camera, Microphone)*

zhangyx1998:
This will make things **MUCH HARDER** for local network application development.

In my case, I want to serve a webpage from a robot which accepts gamepad controls. The point is everyone in the local network should be able to connect to the robot from their own device with **zero configuration** (i.e. self-signed certs are not feasible). Forcing secure context will **destroy** these kind of applications.

Instead of forcing a secure context, can the engine just go ahead and prompt for user permission?

BlobTheKat:
> Instead of forcing a secure context, can the engine just go ahead and prompt for user permission?
> 

I disagree, requiring user permission is a nuisance to the user.

What would make more sense is a general interface for receiving controller input that doesn't expose device-specific details (similar to how keyboard input is exposed exclusively through `KeyboardEvent`s)

If this is implemented, individual controller/gamepad details could still be accessible behind a permission, similar to `getScreenDetails()`

This seems like the most consistent and privacy-preserving approach, not sure how the rationale behind this PR came to be.

zhangyx1998:
Thank you @BlobTheKat for replying. I've opened an issue and adopted your opinions.

Please feel free to correct me if I did not understand you correctly.

frank26080115:
I am running a webserver from a ESP8266, under the ExpressLRS project, so the MCU itself is incapable of encryption and I don't have ways of adding libraries to the project. I am trying to add gamepad functionality to ExpressLRS through it's Wi-Fi soft-AP web server that is delivering a HTML page with Javascript utilizing the gamepad API. It is my hope to use this as a platform to develop low cost remotely controlled educational toy robots.

I saw a console warning that lead me to this pull request. It said there are instructions to resolve the error here. How can I resolve it?

Thanks

partyblob:
> I am running a webserver from a ESP8266, under the ExpressLRS project, so the MCU itself is incapable of encryption and I don't have ways of adding libraries to the project. I am trying to add gamepad functionality to ExpressLRS through it's Wi-Fi soft-AP web server that is delivering a HTML page with Javascript utilizing the gamepad API. It is my hope to use this as a platform to develop low cost remotely controlled educational toy robots.
> 
> I saw a console warning that lead me to this pull request. It said there are instructions to resolve the error here. How can I resolve it?
> 
> Thanks

The decision has been reverted, the requirement has been dropped and the warning should go away in future versions of chrome (See #194 and #203)

PR #123, opened 2019-10-24
Spec liveness of Gamepad and GamepadButton objects (#8)

### LINKED ISSUE #8: Need to spec liveness of Gamepad objects

nondebug:
Closes #8

Implementation commitment:

***
<a href="https://pr-preview.s3.amazonaws.com/nondebug/gamepad/pull/123.html" title="Last updated on Nov 21, 2019, 1:35 AM UTC (f71842d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/123/eb3e7ad...nondebug:f71842d.html" title="Last updated on Nov 21, 2019, 1:35 AM UTC (f71842d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -479,7 +479,7 @@
       <pre class="idl">
         [Exposed=Window]
         partial interface Navigator {
-          sequence&lt;Gamepad?&gt; getGamepads();
+          [NewObject] sequence&lt;Gamepad?&gt; getGamepads();
         };
       </pre>
       <dl>
--- a/index.html
+++ b/index.html
@@ -209,7 +209,11 @@
         <dfn>Gamepad</dfn> interface
       </h2>
       <p>
-        This interface defines an individual gamepad device.
+        This interface represents an individual gamepad device. A Gamepad is a
+        live object; that is, user interaction with the device will be reflected
+        in the state of the object's properties. When the system receives new
+        data from a connected gamepad, a microtask is queued in the current
+        event loop to update the state of the Gamepad object.
       </p>
       <pre class="idl" data-cite="HR-TIME">
         [Exposed=Window]
@@ -230,7 +234,8 @@
         <dd>
           An identification string for the gamepad. This string identifies the
           brand or style of connected gamepad device. Typically, this will
-          include the USB vendor and a product ID.
+          include the USB vendor and a product ID. The ID string MUST NOT change
+          once the Gamepad object has been returned to script.
         </dd>
         <dt>
           <dfn>index</dfn> attribute
@@ -296,19 +301,19 @@
           then Y. It is RECOMMENDED that axes appear in decreasing order of
           importance, such that element 0 and 1 typically represent the X and Y
           axis of a directional stick. The same object MUST be returned until
-          the <a>user agent</a> needs to return different values (or values in
-          a different order).
+          the <a>user agent</a> needs to modify the length of the array or
+          return different axis values.
         </dd>
         <dt>
           <dfn>buttons</dfn> attribute
         </dt>
         <dd>
-          Array of button states for all buttons of the gamepad. It is
+          Array of GamepadButton objects representing all buttons of the gamepad. It is
           RECOMMENDED that buttons appear in decreasing importance such that
           the primary button, secondary button, tertiary button, and so on
           appear as elements 0, 1, 2, ... in the buttons array. The same object
-          MUST be returned until the <a>user agent</a> needs to return
-          different values (or values in a different order).
+          MUST be returned until the <a>user agent</a> needs to modify the
+          length of the array or the button ordering.
         </dd>
       </dl>
     </section>
@@ -318,7 +323,13 @@
       </h2>
       <p>
         This interface defines the state of an individual button on a gamepad
-        device.
+        device. A GamepadButton is a live object; that is, user interaction with
+        the button represented by the object will be reflected in the object's
+        properties. When the system receives new data from a connected gamepad,
+        a microtask is queued in the current event loop to update the state of
+        the GamepadButton object. If a gamepad device is disconnected, the
+        GamepadButton objects representing its buttons MUST continue to return
+        the last state received from the device before disconnection.
       </p>
       <pre class="idl">
         [Exposed=Window]
@@ -416,7 +427,7 @@
           <dfn>getGamepads</dfn>
         </dt>
         <dd>
-          Retrieve a snapshot of the data for the the currently connected and
+          Retrieve an array of Gamepad objects representing currently connected and
           interacted-with gamepads. Gamepads MUST only appear in the list if
           they are currently connected to the <a>user agent</a>, and at least
           one device has been interacted with by the user. If no devices have
--- a/index.html
+++ b/index.html
@@ -291,7 +291,7 @@
           </p>
           <p>
           If no data has been received from the hardware, the value of the
-          <code>timestamp</code> attribute MUST be the time relative to
+          `timestamp` attribute MUST be the time relative to
           <code>navigationStart</code> when the <a>Gamepad</a> object was first
           made available to script.
           </p>
--- a/index.html
+++ b/index.html
@@ -343,7 +343,7 @@
           </p>
           <p>
           Once the system is notified that a previously connected gamepad has
-          been disconnected, the axes attribute MUST continue to return the
+          been disconnected, the {{Gamepad/axes}} attribute MUST continue to return the
           last state received from the gamepad before disconnection.
           </p>
         </dd>
--- a/index.html
+++ b/index.html
@@ -234,8 +234,7 @@
         <dd>
           An identification string for the gamepad. This string identifies the
           brand or style of connected gamepad device. Typically, this will
-          include the USB vendor and a product ID. The ID string MUST NOT change
-          once the Gamepad object has been returned to script.
+          include the USB vendor and a product ID.
         </dd>
         <dt>
           <dfn>index</dfn> attribute
@@ -258,6 +257,30 @@
           lowest previously used index MUST be assigned to the newly connected
           gamepad.
           </p>
+
+          <pre class="example highlight">
+            // Consider three gamepads padA, padB, and padC.
+            // Initially, padA and padB are connected with indices 0 and 1.
+            // padC is not connected.
+            var gamepads = navigator.getGamepads();
+            // The following statements should all evaluate to true.
+            gamepads.length == 2;
+            gamepads[0].index == 0;  // padA
+            gamepads[1].index == 1;  // padB
+            // padA is disconnected.
+            gamepads.length == 2;
+            gamepads[0] == null;
+            gamepads[1].index == 1;  // padB
+            // padC is connected and takes the lowest unassigned index (0).
+            gamepads.length == 2;
+            gamepads[0].index == 0;  // padC
+            gamepads[1].index == 1;  // padB
+            // padA is reconnected and takes the lowest unassigned index (2).
+            gamepads.length == 3;
+            gamepads[0].index == 0;  // padC
+            gamepads[1].index == 1;  // padB
+            gamepads[2].index == 2;  // padA
+          </pre>
         </dd>
         <dt>
           <dfn>connected</dfn> attribute
@@ -379,7 +402,7 @@
       <p>
         This interface represents the current state of an individual button on a gamepad
         device. A {{GamepadButton}} is a live object; that is, user interaction with
-        the button represented by the object will be reflected in the object's
+        the button represented by the object is periodically reflected in the object's
         properties. When the system receives new data from a connected gamepad,
         a microtask MUST be queued with the user interaction task source to
         update the state of the GamepadButton objects representing the buttons
@@ -500,10 +523,6 @@
           {{Gamepad/index}} attribute. Array indices for which there is no
           connected Gamepad with the corresponding index should return null.
 
-          <p>
-          A new array MUST be returned for each call to getGamepads().
-          </p>
-          
           <p>
           The gamepad state returned from getGamepads() does not reflect
           disconnection or connection until after the <a>gamepaddisconnected</a>
--- a/index.html
+++ b/index.html
@@ -323,7 +323,7 @@
       </h2>
       <p>
         This interface defines the state of an individual button on a gamepad
-        device. A GamepadButton is a live object; that is, user interaction with
+        device. A {{GamepadButton}} is a live object; that is, user interaction with
         the button represented by the object will be reflected in the object's
         properties. When the system receives new data from a connected gamepad,
         a microtask is queued in the current event loop to update the state of
--- a/index.html
+++ b/index.html
@@ -210,7 +210,7 @@
       </h2>
       <p>
         This interface represents an individual gamepad device. A Gamepad is a
-        live object; that is, user interaction with the device will be reflected
+        live object; that is, user interaction with the device is periodically reflected
         in the state of the object's properties. When the system receives new
         data from a connected gamepad, a microtask MUST be queued with the user
         interaction task source to update the state of the {{Gamepad}} object.
--- a/index.html
+++ b/index.html
@@ -427,7 +427,7 @@
           <dfn>getGamepads</dfn>
         </dt>
         <dd>
-          Retrieve an array of Gamepad objects representing currently connected and
+          Retrieve a sequence of {{Gamepad}} objects representing currently connected and
           interacted-with gamepads. Gamepads MUST only appear in the list if
           they are currently connected to the <a>user agent</a>, and at least
           one device has been interacted with by the user. If no devices have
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -209,7 +209,11 @@ <h2>
         <dfn>Gamepad</dfn> interface
       </h2>
       <p>
-        This interface defines an individual gamepad device.
+        This interface represents an individual gamepad device. A Gamepad is a
+        live object; that is, user interaction with the device will be reflected
+        in the state of the object's properties. When the system receives new
+        data from a connected gamepad, a microtask is queued in the current
```
marcoscaceres: Can probably drop "in the current event loop" as that's a bit ambiguous. 

I wonder if we should use the "user interaction task source"?

index.html
```diff
@@ -209,7 +209,11 @@ <h2>
...
+        event loop to update the state of the Gamepad object.
```
```suggestion
        event loop to update the state of the {{Gamepad}} object.
```

index.html
```diff
@@ -296,19 +301,19 @@ <h2>
...
-          a different order).
+          the <a>user agent</a> needs to modify the length of the array or
+          return different axis values.
         </dd>
         <dt>
           <dfn>buttons</dfn> attribute
         </dt>
         <dd>
-          Array of button states for all buttons of the gamepad. It is
+          Array of GamepadButton objects representing all buttons of the gamepad. It is
```
```suggestion
          Array of {{GamepadButton}} objects representing all buttons of the gamepad. It is
```

index.html
```diff
@@ -296,19 +301,19 @@ <h2>
...
           RECOMMENDED that buttons appear in decreasing importance such that
           the primary button, secondary button, tertiary button, and so on
           appear as elements 0, 1, 2, ... in the buttons array. The same object
-          MUST be returned until the <a>user agent</a> needs to return
-          different values (or values in a different order).
+          MUST be returned until the <a>user agent</a> needs to modify the
+          length of the array or the button ordering.
```
marcoscaceres: This is somewhat ambiguous... it could be read like, if an axis changes, then replace all every `GamepadButton`.

index.html
```diff
@@ -318,7 +323,13 @@ <h2>
       </h2>
       <p>
         This interface defines the state of an individual button on a gamepad
-        device.
+        device. A GamepadButton is a live object; that is, user interaction with
```
```suggestion
        device. A {{GamepadButton}} is a live object; that is, user interaction with
```

marcoscaceres: I think this should just read:
"A {{GamepadButton}} is a live object: When the system receives new data from a connected gamepad..." 

And drop the "that is, user interaction with....". That makes it more clear that updates to the various properties are dependent on the queued microtasks.

index.html
```diff
@@ -318,7 +323,13 @@ <h2>
...
+        the button represented by the object will be reflected in the object's
+        properties. When the system receives new data from a connected gamepad,
+        a microtask is queued in the current event loop to update the state of
+        the GamepadButton object. If a gamepad device is disconnected, the
```
marcoscaceres: The "If a gamepad device is disconnected..." should probably be its own paragraph. Eventually, we need to turn that also into a queued task... "Upon disconnection of a gamepad, queue a task to... "

index.html
```diff
@@ -416,7 +427,7 @@ <h2>
           <dfn>getGamepads</dfn>
         </dt>
         <dd>
-          Retrieve a snapshot of the data for the the currently connected and
+          Retrieve an array of Gamepad objects representing currently connected and
```
```suggestion
          Retrieve a sequence of {{Gamepad}} objects representing currently connected and
```

marcoscaceres (changes requested):

index.html
nondebug: That seems like the best approach. According to the HTML5 spec, input events should be queued this way:

"Asynchronous events sent in response to user input (e.g. click events) must be fired using tasks queued with the user interaction task source."

https://www.w3.org/TR/2014/REC-html5-20141028/webappapis.html#generic-task-sources

I think specifying the live object updates in terms of the user interaction task source will make it easier to spec gamepad button/axis events in the future.

index.html
nondebug: Good idea. I uploaded a new commit where I've pulled out behavior that should be tied to system events (gamepad connection, gamepad disconnection, gamepad state update). Eventually these should be combined into an algorithm for each system event.

index.html
nondebug: I've rewritten the description of the axes and buttons attributes to make it clearer when the array should be reused and when it should be recreated.

index.html
```diff
@@ -209,7 +209,11 @@ <h2>
         <dfn>Gamepad</dfn> interface
       </h2>
       <p>
-        This interface defines an individual gamepad device.
+        This interface represents an individual gamepad device. A Gamepad is a
+        live object; that is, user interaction with the device will be reflected
```
```suggestion
        live object; that is, user interaction with the device is periodically reflected
```

index.html
```diff
@@ -230,45 +234,75 @@ <h2>
         <dd>
           An identification string for the gamepad. This string identifies the
           brand or style of connected gamepad device. Typically, this will
-          include the USB vendor and a product ID.
+          include the USB vendor and a product ID. The ID string MUST NOT change
+          once the Gamepad object has been returned to script.
```
marcoscaceres: We probably don't need to say (unless this has happened in practice?): 

> The ID string MUST NOT change once the Gamepad object has been returned to script.

As nothing in the spec causes the to change.

index.html
```diff
@@ -230,45 +234,75 @@ <h2>
...
+          assigned to any currently connected gamepad, starting at zero.
+          </p>
+          <p>
+          When the system is notified that a previously connected gamepad has
+          become unavailable, the `index` attribute of other gamepads MUST NOT
+          be reassigned. However, if a gamepad is disconnected, and
+          subsequently the same or a different gamepad is connected, the
+          lowest previously used index MUST be assigned to the newly connected
+          gamepad.
+          </p>
```
marcoscaceres, line 259: Should we add an example for clarity? 

 1. getGamepads() returns - [pad1, pad2]
 1. pad1 disconnects 
 1. getGamepads() returns - [null, pad2]
 1. pad3 connects, event fire
 1. getGamepads() returns - [pad3, pad2]
 1 pad1 re-connects
 1. getGamepads() returns - [pad3, pad2, pad1]

index.html
```diff
@@ -230,45 +234,75 @@ <h2>
...
           hardware. The value must be relative to the
           <code>navigationStart</code> attribute of the
           <a>PerformanceTiming</a> interface. Since values are monotonically
           increasing they can be compared to determine the ordering of updates,
           as newer values will always be greater than or equal to older values.
+          </p>
+          <p>
           If no data has been received from the hardware, the value of the
-          <code>timestamp</code> attribute should be the time relative to
+          <code>timestamp</code> attribute MUST be the time relative to
```
```suggestion
          `timestamp` attribute MUST be the time relative to
```

index.html
```diff
@@ -295,20 +330,45 @@ <h2>
...
+          microtask MUST be queued with the user interaction task source to
+          update the gamepad state. If the gamepad has the same axes in the same
+          order as the previous update, and if no axis values have changed
+          since the previous update, then the axes attribute MUST return the
+          same array object. If the number of axes has changed, or if any axis
+          value has changed, then the axis attribute MUST return a new array.
+          </p>
+          <p>
+          Once the system is notified that a previously connected gamepad has
+          been disconnected, the axes attribute MUST continue to return the
```
```suggestion
          been disconnected, the {{Gamepad/axes}} attribute MUST continue to return the
```

index.html
```diff
@@ -295,20 +330,45 @@ <h2>
...
+          button, tertiary button, and so on appear as elements 0, 1, 2, ...
+          in the buttons array.
+          </p>
+          <p>
+          When the system receives new data from a connected gamepad, a
+          microtask MUST be queued with the user interaction task source to
+          update the gamepad state. If the gamepad has the same buttons in the
+          same order as the previous update, the buttons attribute MUST return
+          the same array object. If the number of buttons or the ordering of
+          buttons has changed, the buttons attribute MUST return a new array
```
marcoscaceres, line 390: Oh, this would break  JS invariant: 

```
a.foo === a.foo
```

The same array must always be returned... just the objects inside it might have changed.

index.html
```diff
@@ -295,20 +330,45 @@ <h2>
...
+          object. If a button in the new array represents the same physical
+          button as a button from the previous array, the same GamepadButton
```
```suggestion
          button as a button from the previous array, the same {{GamepadButton}}
```

index.html
```diff
@@ -428,6 +499,10 @@ <h2>
           each Gamepad present at the index in the array specified by its
           {{Gamepad/index}} attribute. Array indices for which there is no
           connected Gamepad with the corresponding index should return null.
+
+          <p>
+          A new array MUST be returned for each call to getGamepads().
```
marcoscaceres: We can delete this and annotate getGamepads() with [`[NewObject]`](https://heycam.github.io/webidl/#NewObject).

marcoscaceres (changes requested):
Couple of issues... the button attribute returning a new object sometimes might be a problem, so we should figure something out there.

index.html
nondebug: This suggests that we can't make Gamepad a live object and allow the number of buttons or axes to change. This would diverge from the original spec, but is probably not important since I don't think there are any implementations that support this behavior. I'll remove the language about changes to the length of the buttons and axes arrays.

nondebug: This would work for the buttons attribute but I'm not sure how to resolve it for axes. The axes attribute is FrozenArray<double> so we can't change the values in-place like we can for the buttons array.

I tested the current behavior in Firefox. `pad.axes === pad.axes` always returns true, but if you save the axes array before and after interacting with the joystick then you'll get different arrays:

```
let pad = navigator.getGamepads()[0];
let axes = pad.axes;
axes === pad.axes; // true
// wiggle the thumbstick
axes === pad.axes; // false
```

index.html
nondebug: I added the example but renamed pad1,pad2,pad3 -> padA,padB,padC for clarity

index.html
nondebug: Sure, removed.

index.html
marcoscaceres: argh... that's not good. @domenic, sorry to drag you in... if you have some cycles, could you suggest how we might proceed? This is a bit a of a pickle :( 

We might need an array-like object here instead 🤢

domenic: The behavior defined in the previous comment seems right to me. There is an async turn boundary between the value of the property changing which is very reasonable.

Also, I suspect both buttons and axes should have this behavior, not just axes.

marcoscaceres: I guess that make sense... it doesn't feel good to swap out arrays, but yes, the async/looping aspects does make it reasonable. Thank you.

marcoscaceres:
@kearwood and @toji just a follow up from TPAC... here is the initial specification of live objects. We want to get your input/review from a WebXR perspective.

PR #140, opened 2020-09-15
chore(ci): use the spec-prod GitHub action

sidvishnoi:
We plan to deprecate [respec-w3c-auto-publish](https://github.com/w3c/respec-w3c-auto-publish) action in favor of a more generic [`spec-prod`](https://github.com/sidvishnoi/spec-prod/) action. The newer action is more consistent and requires fewer inputs.

It also supports deploying generated static ReSpec document to GitHub pages, but we probably don't want to switch to that here yet, as ReSpec runtime has some nice features.


sidvishnoi:

marcoscaceres (approved):

PR #143, opened 2020-10-29
Add gamepad light indicator extension

### LINKED ISSUE #67: Light Indicator
Many controllers support the concept of a light indicator to represent the player index of a controller. This is helpful to the user to know which player character is being controlled without testing each controller.  Additionally, some games are only playable using a controller with a specific player index. In the context of VR, controllers may support a light indicator to assist the application in determining the location of the player. Light indicators range from monochrome light indicators (ON/OFF) to RGB light indicators.  

We'd like to bring gamepad extension support to the same level of support provided by native functionality with the following proposal: [Light Indicator Proposal](https://github.com/knyg/gamepad/commit/cafe2dd405038d9dc25d852d1c4a61b03b3381db)

Please let us know if you have any questions, comments, or feedback.

fernando-80:
Closes #67

Implementation commitment:

 * [ ] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=1523351)

### DIFF
--- a/extensions.html
+++ b/extensions.html
@@ -377,7 +377,7 @@
       </pre>
       
       <dl data-dfn-for="GamepadLightColor">
-        <dt><dfn>red</dt></dfn>
+        <dt><dfn>red</dt> member</dfn>
         <dd>
           Red component of the light color, or non-zero value for an
           <code>on-off</code> light indicator that indicates ON.
--- a/extensions.html
+++ b/extensions.html
@@ -450,8 +450,7 @@
           readonly attribute GamepadHand hand;
           readonly attribute FrozenArray&lt;GamepadHapticActuator> hapticActuators;
           readonly attribute GamepadPose? pose;
-          readonly attribute FrozenArray&lt;GamepadLightIndicator>?
-          lightIndicators;
+          readonly attribute FrozenArray&lt;GamepadLightIndicator>? lightIndicators;
         };
       </pre>
 
--- a/extensions.html
+++ b/extensions.html
@@ -383,7 +383,7 @@
           <code>on-off</code> light indicator that indicates ON.
         </dd>
         
-        <dt><dfn>green</dt></dfn>
+        <dt><dfn>green</dt> member</dfn>
         <dd>
           Blue component of the light color, or non-zero value for an
           <code>on-off</code> light indicator that indicates ON.
--- a/extensions.html
+++ b/extensions.html
@@ -370,9 +370,9 @@
     
       <pre class="idl">
         dictionary GamepadLightColor {
-          required octet red;
-          required octet green;
-          required octet blue;
+          octet red = 0;
+          octet green = 0;
+          octet blue = 0;
        };
       </pre>
       
--- a/extensions.html
+++ b/extensions.html
@@ -334,6 +334,109 @@
         </dd>
       </dl>
     </section>
+    
+    <section>
+      <h2><dfn>GamepadLightIndicatorType</dfn> Enum</h2>
+      <p>
+        This enum defines the type of light indicators supported by gamepads.
+      </p>
+      
+      <pre class="idl">
+        enum GamepadLightIndicatorType {
+          "on-off",
+          "rgb"
+        };
+      </pre>
+      
+      <dl data-dfn-for="GamepadLightIndicatorType">
+        <dt>"<dfn>on-off</dfn>"</dt>
+        <dd>
+          Light indicator that supports a single color.
+        </dd>
+        
+        <dt>"<dfn>rgb</dfn>"</dt>
+        <dd>
+          Light indicator that supports multiple colors.
+        </dd>
+      </dl>
+    </section>
+    
+    <section>
+      <h2><dfn>GamepadLightColor</dfn></h2>
+      <p>
+        This represents the color of a light indicator. The default value should
+        be set to 0 for red, green, and blue.
+      </p>
+    
+      <pre class="idl">
+        dictionary GamepadLightColor {
+          required octet red;
+          required octet green;
+          required octet blue;
+       };
+      </pre>
+      
+      <dl data-dfn-for="GamepadLightColor">
+        <dt><dfn>red</dt></dfn>
+        <dd>
+          Red component of the light color, or non-zero value for an
+          <code>on-off</code> light indicator that indicates ON.
+        </dd>
+        
+        <dt><dfn>green</dt></dfn>
+        <dd>
+          Blue component of the light color, or non-zero value for an
+          <code>on-off</code> light indicator that indicates ON.
+        </dd>
+        
+        <dt><dfn>blue</dt></dfn>
+        <dd>
+          Green component of the light color, or non-zero value for an
+          <code>on-off</code> light indicator that indicates ON.
+        </dd>
+      </dl>
+    </section>
+    
+    <section>
+      <h2><dfn>GamepadLightIndicator</dfn></h2>
+      <p>
+        This interface defines a light indicator.
+      </p>
+      
+      <pre class="idl">
+        [Exposed=Window, SecureContext]
+        interface GamepadLightIndicator {
+          readonly attribute GamepadLightIndicatorType type;
+        
+          Promise&lt;long&gt; setColor(GamepadLightColor color);
+      };
+      </pre>
+      
+      <dl data-dfn-for="GamepadLightIndicator">
+        <dt><dfn>type</dt></dfn>
+        <dd>
+          Type of light indicator supported by the gamepad.
+        </dd>
+        
+        <dt><dfn>setColor</dt></dfn>
+        <dd>
+          Sets the color of a light indicator, or sets the light indicator to ON
+          or OFF.
+          
+          For <code>on-off</code> light indicators, one or more non zero values
+          denotes the light indicator is ON, whereas a zero value for all
+          components indicates the light indicator is OFF.
+          
+          For RGB light indicators, one or more non zero values specify the
+          color of the light indicator, whereas all zero values indicate the
+          light indicator is OFF.
+          
+          The returned Promise will resolve <code>true</code> when setting the
+          color has succeeded, and will reject the device API error code on
+          failure.
+        </dd>
+      </dl>
+    </section>
 
     <section>
       <h2>Partial <dfn>Gamepad</dfn> Interface</h2>
@@ -347,6 +450,8 @@
           readonly attribute GamepadHand hand;
           readonly attribute FrozenArray&lt;GamepadHapticActuator> hapticActuators;
           readonly attribute GamepadPose? pose;
+          readonly attribute FrozenArray&lt;GamepadLightIndicator>?
+          lightIndicators;
         };
       </pre>
 
@@ -370,6 +475,12 @@
           Includes position, orientation, velocity, and acceleration. If the
           hardware cannot supply any pose values, MUST be set to
           <code>null</code>.
+        </dd>
+        
+        <dt><dfn>lightIndicators</dfn></dt>
+        <dd>
+          A list of all light indicators in the gamepad. <code>null</code> if
+          the gamepad does not have or does not support a light indicator.
         </dd>
       </dl>
     </section>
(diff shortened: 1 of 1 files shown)


fernando-80:
@nondebug @knyg @sagoston light-indicator extension PR ^^^

extensions.html
```diff
@@ -334,6 +334,109 @@ <h2><dfn>GamepadPose</dfn> Interface</h2>
...
+          denotes the light indicator is ON, whereas a zero value for all
+          components indicates the light indicator is OFF.
+          
+          For RGB light indicators, one or more non zero values specify the
+          color of the light indicator, whereas all zero values indicate the
+          light indicator is OFF.
+          
+          The returned Promise will resolve <code>true</code> when setting the
+          color has succeeded, and will reject the device API error code on
+          failure.
```
```suggestion
          <p>
          Sets the color of a light indicator, or sets the light indicator to ON
          or OFF.
          </p>
          
          <p>
          For <code>on-off</code> light indicators, one or more non zero values
          denotes the light indicator is ON, whereas a zero value for all
          components indicates the light indicator is OFF.
          </p>
          
          <p>
          For RGB light indicators, one or more non zero values specify the
          color of the light indicator, whereas all zero values indicate the
          light indicator is OFF.
          </p>
          
          <p>
          The returned Promise will resolve <code>true</code> when setting the
          color has succeeded, and will reject the device API error code on
          failure.
          </p>
```

extensions.html
```diff
@@ -347,6 +450,8 @@ <h2>Partial <dfn>Gamepad</dfn> Interface</h2>
           readonly attribute GamepadHand hand;
           readonly attribute FrozenArray&lt;GamepadHapticActuator> hapticActuators;
           readonly attribute GamepadPose? pose;
+          readonly attribute FrozenArray&lt;GamepadLightIndicator>?
+          lightIndicators;
```
```suggestion
          readonly attribute FrozenArray&lt;GamepadLightIndicator>? lightIndicators;
```

sagoston (approved):

sagoston (approved):

marcoscaceres:
Hi @fernando-80, could you kindly please [join the WebApps working group](https://www.w3.org/2004/01/pp-impl/114929/join)? That will make the IPR bot happy.

extensions.html
```diff
@@ -334,6 +334,117 @@ <h2><dfn>GamepadPose</dfn> Interface</h2>
...
+        This represents the color of a light indicator. The default value should
+        be set to 0 for red, green, and blue.
+      </p>
+    
+      <pre class="idl">
+        dictionary GamepadLightColor {
+          required octet red;
+          required octet green;
+          required octet blue;
+       };
```
marcoscaceres: I wonder if these should just default to 0?

extensions.html
```diff
@@ -334,6 +334,117 @@ <h2><dfn>GamepadPose</dfn> Interface</h2>
...
+      };
+      </pre>
+      
+      <dl data-dfn-for="GamepadLightIndicator">
+        <dt><dfn>type</dt></dfn>
+        <dd>
+          Type of light indicator supported by the gamepad.
+        </dd>
+        
+        <dt><dfn>setColor</dt></dfn>
```
```suggestion
        <dt><dfn>setColor()</dt> method</dfn>
```

extensions.html
```diff
@@ -334,6 +334,117 @@ <h2><dfn>GamepadPose</dfn> Interface</h2>
...
+        <dd>
+          <p>
+          Sets the color of a light indicator, or sets the light indicator to ON
+          or OFF.
+          </p>
+          
+          <p>
+          For <code>on-off</code> light indicators, one or more non zero values
+          denotes the light indicator is ON, whereas a zero value for all
+          components indicates the light indicator is OFF.
```
```suggestion
          denotes the light indicator is on, whereas a zero value for all
          components indicates the light indicator is off.
```

extensions.html
```diff
@@ -397,6 +508,7 @@ <h2>Partial <dfn>Gamepad</dfn> Interface</h2>
           readonly attribute GamepadHand hand;
           readonly attribute FrozenArray&lt;GamepadHapticActuator> hapticActuators;
           readonly attribute GamepadPose? pose;
+          readonly attribute FrozenArray&lt;GamepadLightIndicator>? lightIndicators;
```
marcoscaceres: Is it usual for there to be multiple light indicators?

marcoscaceres: Better to just make this an empty array. Then there is less type switching. 
```suggestion
          readonly attribute FrozenArray&lt;GamepadLightIndicator> lightIndicators;
```

extensions.html
```diff
@@ -423,6 +535,11 @@ <h2>Partial <dfn>Gamepad</dfn> Interface</h2>
           <code>null</code>.
         </dd>
         
+        <dt><dfn>lightIndicators</dfn></dt>
+        <dd>
+          A list of all light indicators in the gamepad. <code>null</code> if
+          the gamepad does not have or does not support a light indicator.
```
marcoscaceres: As above... 

```suggestion
          A list of all light indicators in the gamepad. The array is empty if
          the gamepad does not have, or does not support, a light indicator.
```

extensions.html
```suggestion
        dictionary GamepadLightColor {
          octet red = 0;
          octet green = 0;
          octet blue = 0;
       };
```

extensions.html
```diff
@@ -334,6 +334,117 @@ <h2><dfn>GamepadPose</dfn> Interface</h2>
...
+          </p>
+          
+          <p>
+          For RGB light indicators, one or more non zero values specify the
+          color of the light indicator, whereas all zero values indicate the
+          light indicator is OFF.
+          </p>
+          
+          <p>
+          The returned Promise will resolve <code>true</code> when setting the
```
marcoscaceres, line 440: I don't know if this can be guaranteed to succeed. @nondebug, do you have an opinion about this?

extensions.html
```diff
@@ -334,6 +334,117 @@ <h2><dfn>GamepadPose</dfn> Interface</h2>
...
+          color has succeeded, and will reject the device API error code on
+          failure.
```
marcoscaceres, line 442: This needs to reject with a an actual DOM Exception of some kind. I think this is overly broad/underspecified tho...

extensions.html
```diff
@@ -334,6 +334,117 @@ <h2><dfn>GamepadPose</dfn> Interface</h2>
...
+      <pre class="idl">
+        dictionary GamepadLightColor {
+          required octet red;
+          required octet green;
+          required octet blue;
+       };
+      </pre>
+      
+      <dl data-dfn-for="GamepadLightColor">
+        <dt><dfn>red</dt></dfn>
```
```suggestion
        <dt><dfn>red</dt> member</dfn>
```

extensions.html
```diff
@@ -334,6 +334,117 @@ <h2><dfn>GamepadPose</dfn> Interface</h2>
...
+        <dd>
+          Red component of the light color, or non-zero value for an
+          <code>on-off</code> light indicator that indicates ON.
+        </dd>
+        
+        <dt><dfn>green</dt></dfn>
```
```suggestion
        <dt><dfn>green</dt> member</dfn>
```

extensions.html
```diff
@@ -334,6 +334,117 @@ <h2><dfn>GamepadPose</dfn> Interface</h2>
...
+        <dd>
+          Blue component of the light color, or non-zero value for an
+          <code>on-off</code> light indicator that indicates ON.
+        </dd>
+        
+        <dt><dfn>blue</dt></dfn>
```
```suggestion
        <dt><dfn>blue</dt> member</dfn>
```

marcoscaceres (changes requested):
Thanks @fernando-80! Some suggestions...

extensions.html
nondebug: Not for Playstation controllers, but Xbox 360, Wiimote, Joy-Cons, etc have four player indicator LEDs.

extensions.html
nondebug: We typically don't get a signal that the color was changed successfully so we shouldn't claim that in the spec. I think we should reword this so we resolve when the LED color change command is sent and reject if the command couldn't be sent.

extensions.html
```diff
@@ -334,6 +334,117 @@ <h2><dfn>GamepadPose</dfn> Interface</h2>
...
+      <p>
+        This interface defines a light indicator.
+      </p>
+      
+      <pre class="idl">
+        [Exposed=Window, SecureContext]
+        interface GamepadLightIndicator {
+          readonly attribute GamepadLightIndicatorType type;
+        
+          Promise&lt;long&gt; setColor(GamepadLightColor color);
```
marcoscaceres: Given the discussion below, the promise should probably just resolve with `undefined`. 

```suggestion
          Promise&lt;undefined&gt; setColor(GamepadLightColor color);
```

extensions.html
marcoscaceres: Ok, so yeah... it shouldn't resolve with `true` either. 

We probably need to rewrite this whole section. I'd suggest having a look at:
https://w3c.github.io/web-share/#share-method

Basically, we need to do a few checks that can cause the promise to reject... the controller has been disconnected/no longer available, etc. There might be a SecurityError if the permission has been changed before `.setColor()` is called. Also, what should happen if:

```JS
while (true) {
   x.setColor(randomRGB());
}
```

Rate limit? Drop requests on the floor?

extensions.html
```diff
@@ -334,6 +334,117 @@ <h2><dfn>GamepadPose</dfn> Interface</h2>
...
+          Light indicator that supports multiple colors.
+        </dd>
+      </dl>
+    </section>
+    
+    <section>
+      <h2><dfn>GamepadLightColor</dfn></h2>
+      <p>
+        This represents the color of a light indicator. The default value should
+        be set to 0 for red, green, and blue.
```
marcoscaceres: Don't need the description, as default is already state in the IDL. 
```suggestion
        This represents the color of a light indicator.
```

extensions.html
```diff
@@ -334,6 +334,117 @@ <h2><dfn>GamepadPose</dfn> Interface</h2>
...
+        
+        <dt>"<dfn>rgb</dfn>"</dt>
+        <dd>
+          Light indicator that supports multiple colors.
+        </dd>
+      </dl>
+    </section>
+    
+    <section>
+      <h2><dfn>GamepadLightColor</dfn></h2>
```
```suggestion
      <h2><dfn>GamepadLightColor</dfn> dictionary</h2>
```

extensions.html
```diff
@@ -423,6 +535,11 @@ <h2>Partial <dfn>Gamepad</dfn> Interface</h2>
           <code>null</code>.
         </dd>
         
+        <dt><dfn>lightIndicators</dfn></dt>
+        <dd>
+          A list of all light indicators in the gamepad. The array is empty if
+          the gamepad does not have, or does not support, a light indicator.
+
```
```suggestion
          the gamepad does not have, or does not support, a light indicator.
       </dd>
```

extensions.html
```diff
@@ -334,6 +334,116 @@ <h2><dfn>GamepadPose</dfn> Interface</h2>
...
+          </p>
+          
+          <p>
+          For RGB light indicators, one or more non zero values specify the
+          color of the light indicator, whereas all zero values indicate the
+          light indicator is OFF.
+          </p>
+          
+          <p>
+          The returned Promise will resolve <code>true</code> when setting the
```
marcoscaceres, line 440:
```suggestion
          The returned Promise will resolve `undefined` when setting the
```

extensions.html
```diff
@@ -334,6 +334,116 @@ <h2><dfn>GamepadPose</dfn> Interface</h2>
...
+    </section>
+    
+    <section>
+      <h2><dfn>GamepadLightColor</dfn> dictionary</h2>
+      <p>
+        This represents the color of a light indicator.
+      </p>
+    
+      <pre class="idl">
+        dictionary GamepadLightColor {
```
marcoscaceres, line 371: We should investigate what existing color objects we could use from the platform here.

extensions.html
nondebug: We discussed this and couldn't come up with any compelling reasons to keep the Promise. It's unlikely that we'll receive any error indication from the device and there isn't any way we can lose permission for a Gamepad. Let's remove the promise and treat setColor as fire-and-forget.

willuhmjs (approved):
Looks good, excited for this feature.

PR #146, opened 2021-03-12
Add the xr-standard mapping type

toji:
This was reported in https://github.com/immersive-web/webxr-gamepads-module/issues/35

This property was already defined in the [WebXR Gamepad Module](https://immersive-web.github.io/webxr-gamepads-module/#xr-standard-gamepad-mapping), but since WebIDL does not allow for partial enums we ended up repeating the enum with the new value in that spec. This has apparently caused some tooling issues as there are now two conflicting copies of the same enum within the W3C spec library.

I would love it if there was a way to keep this value better scoped to the relevant spec, but lacking that it seems best to lift this up into the main gamepad spec with links off to the definition of the mapping in the WebXR spec and a note that it is reserved for use by that spec.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/pull/146.html" title="Last updated on Mar 17, 2021, 1:33 AM UTC (8bd4404)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/146/e84cb46...8bd4404.html" title="Last updated on Mar 17, 2021, 1:33 AM UTC (8bd4404)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -359,6 +359,7 @@
         enum GamepadMappingType {
           "",
           "standard",
+          "xr-standard",
         };
       </pre>
       <dl>
@@ -375,6 +376,18 @@
         <dd>
           The Gamepad's controls have been mapped to the <a href=
           "#remapping">Standard Gamepad layout</a>.
+        </dd>
+        <dt>
+          <dfn>xr-standard</dfn>
+        </dt>
+        <dd>
+          The Gamepad's controls have been mapped to the <a href=
+          "https://immersive-web.github.io/webxr-gamepads-module/#xr-standard-gamepad-mapping">
+          XR Standard Gamepad Layout</a>. This mapping is reserved for use by
+          the <a href="https://immersive-web.github.io/webxr-gamepads-module">
+          WebXR Gamepad Module</a>. Gamepads returned by
+          {{Navigator/getGamepads()}} MUST NOT report a <a>mapping</a> of
+          {{GamepadMappingType["xr-standard"]}}.
         </dd>
       </dl>
     </section>


nondebug (approved):
This looks good to me.

sagoston (approved):
lgtm

index.html
```diff
@@ -376,6 +377,18 @@ <h2>
           The Gamepad's controls have been mapped to the <a href=
           "#remapping">Standard Gamepad layout</a>.
         </dd>
+        <dt>
+          <dfn>xr-standard</dfn>
+        </dt>
+        <dd>
+          The Gamepad's controls have been mapped to the <a href=
+          "https://immersive-web.github.io/webxr-gamepads-module/#xr-standard-gamepad-mapping">
+          XR Standard Gamepad Layout</a>. This mapping is reserved for use by
```
marcoscaceres, line 386: @toji, in webxr-gamepads-module, would you mind `<dfn export>`ing "XR Standard Gamepad Layout"? Then we can link to it like so:  

```suggestion
          The Gamepad's controls have been mapped to the [=XR Standard
          Gamepad Layout=]. This mapping is reserved for use by
```

index.html
```diff
@@ -376,6 +377,18 @@ <h2>
           The Gamepad's controls have been mapped to the <a href=
           "#remapping">Standard Gamepad layout</a>.
         </dd>
+        <dt>
+          <dfn>xr-standard</dfn>
+        </dt>
+        <dd>
+          The Gamepad's controls have been mapped to the <a href=
+          "https://immersive-web.github.io/webxr-gamepads-module/#xr-standard-gamepad-mapping">
```
himorin, line 385: We'd be better to use /TR/ reference than linking to a draft at github.io.

dontcallmedom:
https://github.com/heycam/webidl/issues/184 is where partial enums have been proposed for reference

index.html
marcoscaceres: Should happen automatically once the definition is exported and this document is republished.

index.html
toji: Added the export to https://github.com/immersive-web/webxr-gamepads-module/pull/36, but we were waiting to land that until this change is merged so that we don't lose the enum definition, even temporarily.

toji:
I believe all the outstanding comments have been addressed at this point. Anything else that should be looked at before merging?

marcoscaceres:
Ok to merge this and we can deal with the hard links after... I need to check why the bots are stuck tho.

marcoscaceres:
Ok, merged. @toji, let me know once stuff on your side gets merged and I'll fix up the links.

toji:
Thanks! Just merged https://github.com/immersive-web/webxr-gamepads-module/pull/36

marcoscaceres:
Awesome, thanks for the update!

Issue #149, opened 2021-05-19
gamepadconnected/disconnected events & non-"fully active" documents


rakina:
Currently the [gamepadconnected](https://w3c.github.io/gamepad/#event-gamepadconnected) and [gamepaddisconnected](https://w3c.github.io/gamepad/#event-gamepaddisconnected) events fire on the (active document's?) window. It's a bit unclear how it interacts with non-"[fully active](https://html.spec.whatwg.org/multipage/browsers.html#fully-active)" (bfcached) documents. I think it's better if the text specify that the window the event is fired to is the [active document](https://html.spec.whatwg.org/multipage/browsers.html#active-document)'s window.

Also, it's probably a good chance to specify how bfcache restores are handled. Currently there's a text that says the `gamepadconnected` should fire "if a gamepad was already connected when the page was loaded". Does this include bfcache restore? What if a gamepad is connected then disconnected (possibly multiple times) while the document is in bfcache? Probably it makes sense to just send the last event (if it's different than the last dispatched event to the document) on bfcache restore (see our WIP [guide](https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/317/b295eb7...rakina:7e7196c.html#gate-fully-active)).
 
cc @domenic @fergal

marcoscaceres:
@rakina, should be ok to add, but I'm wondering if this doesn't apply generally to most (all?) events that get fired at the Window object? I'm wondering if maybe DOM should be providing some guidance here too? cc @annevk

rakina:
Not sure if I get what you're suggesting correctly, but I think this applies to most events that come from "external sources" (e.g. devices), while "internal events" (changes to DOM?) are less of a problem because we won't run tasks/scripts within the non-fully active document itself. Providing general guidance on top of the TAG docs sounds good, how should we do that?

marcoscaceres:
> Not sure if I get what you're suggesting correctly, but I think this applies to most events that come from "external sources" (e.g. devices), 

Right, yes.  

> while "internal events" (changes to DOM?) are less of a problem because we won't run tasks/scripts within the non-fully active document itself. 

In the sentence above, who is "we" in "we won't run tasks"? Do you mean the browser or something specific that guards against that in a spec? 

> Providing general guidance on top of the TAG docs sounds good, how should we do that?

I guess we could have a go here (or in Geo) at preparing a pull request... then based on that, we can see if some general guidance is needed about how to write whatever prose is needed. 

[...]

PR #151, opened 2021-06-10
Add algorithms for getGamepads and events

### LINKED ISSUE #114: Rewrite things as algorithms...
The spec currently lacks algorithms/steps for things, which makes the spec hard to test and to determine what happens when, and in what order. `getGamepads()` in particular should be an algorithm.

nondebug:
Closes #114

Implementation commitment:

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/pull/151.html" title="Last updated on Jul 8, 2021, 11:50 PM UTC (0abe48e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/151/a1fdef2...0abe48e.html" title="Last updated on Jul 8, 2021, 11:50 PM UTC (0abe48e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -161,7 +161,7 @@
         Instances of {{Gamepad}} are created with the internal slots described
         in the following table:
       </p>
-      <table class="simple" dfn-for="Gamepad">
+      <table class="simple">
         <tr>
           <th>
             Internal slot
@@ -203,8 +203,8 @@
             An empty [=sequence=]
           </td>
           <td>
-            A [=sequence=] of `double` values representing the current state of
-            axes exposed by this device
+            A [=sequence=] of {{double}} values representing the current state
+            of axes exposed by this device
           </td>
         </tr>
         <tr>
@@ -356,13 +356,13 @@
         <dd>
           <p>
             The {{Gamepad/timestamp}} allows the author to determine the last
-            time the {{Gamepad/axes}} or {{Gamepad/buttons}} attributes for
-            this gamepad was updated. The value MUST be set to the [=current
-            high resolution time=] each time the system [=receives new button
-            or axis input values=] from the device. If no data has been
-            received from the hardware, {{Gamepad/timestamp}} MUST be the
-            [=current high resolution time=] at the time when the {{Gamepad}}
-            was first made available to script.
+            time the {{Gamepad/axes}} or {{Gamepad/buttons}} attribute for this
+            gamepad was updated. The value MUST be set to the [=current high
+            resolution time=] each time the system [=receives new button or
+            axis input values=] from the device. If no data has been received
+            from the hardware, {{Gamepad/timestamp}} MUST be the [=current high
+            resolution time=] at the time when the {{Gamepad}} was first made
+            available to script.
           </p>
           <p class="warning">
             [=User agent=]s SHOULD set a minimum resolution of |gamepad|'s
@@ -458,9 +458,8 @@
           <li>Let |gamepad:Gamepad| be the {{Gamepad}} object representing the
           device that received new button or axis input values.
           </li>
-          <li>[=Queue a global task=] on the [=relevant global object=] of
-          |gamepad| using the [=gamepad task source=] to [=update gamepad
-          state=] for |gamepad|.
+          <li>[=Queue a task=] on the [=gamepad task source=] to [=update
+          gamepad state=] for |gamepad|.
           </li>
         </ol>
         <p>
@@ -497,9 +496,12 @@
                       <li>Set |connectedGamepad|.{{Gamepad/[[timestamp]]}} to
                       |now|.
                       </li>
-                      <li>If |gamepad|'s [=relevant global object=] is a
-                      {{Window}} object, [=queue a global task=] on |gamepad|'s
-                      [=relevant global object=] using the [=gamepad task
+                      <li>Let |document:Document?| be |gamepad|'s [=relevant
+                      global object=]'s [=associated `Document`=]; otherwise
+                      `null`.
+                      </li>
+                      <li>If |document| is not `null` and is [=Document/fully
+                      active=], then [=queue a task=] on the [=gamepad task
                       source=] to [=fire an event=] named {{gamepadconnected}}
                       at |gamepad|'s [=relevant global object=] using
                       {{GamepadEvent}} with its {{GamepadEvent/gamepad}}
@@ -517,7 +519,7 @@
           following steps:
         </p>
         <ol>
-          <li>Let |axisValues:list| be a [=list=] of `unsigned long` values
+          <li>Let |axisValues:list| be a [=list=] of {{unsigned long}} values
           representing the most recent logical axis input values for each axis
           input of the device represented by |gamepad|.
           </li>
@@ -553,7 +555,7 @@
           the following steps:
         </p>
         <ol>
-          <li>Let |buttonValues:list| be a [=list=] of `unsigned long` values
+          <li>Let |buttonValues:list| be a [=list=] of {{unsigned long}} values
           representing the most recent logical button input values for each
           button input of the device represented by |gamepad|.
           </li>
@@ -682,11 +684,11 @@
           device represented by |gamepad|.
           </li>
           <li>Set |gamepad|.{{Gamepad/[[axisMinimums]]}} to a [=list=] of
-          `unsigned long` values with [=list/size=] equal to |inputCount|
+          {{unsigned long}} values with [=list/size=] equal to |inputCount|
           containing minimum logical values for each of the axis inputs.
           </li>
           <li>Set |gamepad|.{{Gamepad/[[axisMaximums]]}} to a [=list=] of
-          `unsigned long` values with [=list/size=] equal to |inputCount|
+          {{unsigned long}} values with [=list/size=] equal to |inputCount|
           containing maximum logical values for each of the axis inputs.
           </li>
           <li>Initialize |unmappedInputList| to be an empty [=list=].
@@ -767,11 +769,11 @@
           the device represented by |gamepad|.
           </li>
           <li>Set |gamepad|.{{Gamepad/[[buttonMinimums]]}} to be a [=list=] of
-          `unsigned long` values with [=list/size=] equal to |inputCount|
+          {{unsigned long}} values with [=list/size=] equal to |inputCount|
           containing minimum logical values for each of the button inputs.
           </li>
           <li>Set |gamepad|.{{Gamepad/[[buttonMaximums]]}} to be a [=list=] of
-          `unsigned long` values with [=list/size=] equal to |inputCount|
+          {{unsigned long}} values with [=list/size=] equal to |inputCount|
           containing maximum logical values for each of the button inputs.
           </li>
           <li>Initialize |unmappedInputList| to be an empty [=list=].
@@ -847,24 +849,6 @@
           <li>Return |buttons|.
           </li>
         </ol>
-        <p>
-          User agents implementing this specification must provide a new DOM
-          event, named {{gamepadconnected}}. The corresponding event MUST be of
-          type {{GamepadEvent}} and, if [=allowed to use=] the `"gamepad"`
-          permission, MUST fire on the {{Window}} object. Registration for and
-          firing of the {{gamepadconnected}} event MUST follow the usual
-          behavior of DOM Events. [[DOM]]
-        </p>
-        <p>
-          A [=user agent=] MUST dispatch this event type to indicate the user
-          has connected a gamepad. If a gamepad was already connected when the
-          page was loaded, the {{gamepadconnected}} event SHOULD be dispatched
-          when the user presses a button or moves an axis.
-        </p>
-        <p>
-          A [=user agent=] MUST NOT dispatch this event type if the
-          [=environment settings object=] is a [=non-secure context=].
-        </p>
       </section>
     </section>
     <section data-dfn-for="GamepadButton" data-link-for="GamepadButton">
@@ -887,7 +871,7 @@
         Instances of {{GamepadButton}} are created with the internal slots
         described in the following table:
       </p>
-      <table class="simple" dfn-for="GamepadButton">
+      <table class="simple">
         <tr>
           <th>
             Internal slot
@@ -929,7 +913,7 @@
             0.0
           </td>
           <td>
-            A `double` representing the button value scaled to the range [0.0
+            A {{double}} representing the button value scaled to the range [0.0
             .. 1.0]
           </td>
         </tr>
@@ -1056,7 +1040,7 @@
         Instances of {{Navigator}} are created with the internal slots
         described in the following table:
       </p>
-      <table class="simple" dfn-for="Navigator">
+      <table class="simple">
         <tr>
           <th>
             Internal slot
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -170,6 +170,43 @@ <h2>
...
+          <th>Internal slot
+          <th>Initial value
+          <th>Description (non-normative)
+        <tr>
+          <td><dfn>[[\exposed]]</dfn>
+          <td>`false`
+          <td>A flag indicating that the {{Gamepad}} object has been exposed to script.
+        <tr>
+          <td><dfn>[[\axisMapping]]</dfn>
+          <td>`{}`
```
marcoscaceres: In the internal slots, we can switch to using infra types. That gives us nicer ways of working with things. 

```suggestion
          <td>An empty [=ordered map=]. 
```

index.html
```diff
@@ -170,6 +170,43 @@ <h2>
...
+          <td>Mapping from unmapped axis index to an index in the {{Gamepad/axes}} array
+        <tr>
+          <td><dfn>[[\axisMinimums]]</dfn>
+          <td>`null`
+          <td>An array containing the minimum logical value for each axis.
+        <tr>
+          <td><dfn>[[\axisMaximums]]</dfn>
+          <td>`null`
+          <td>An array containing the maximum logical value for each axis.
```
marcoscaceres: Probably want a list or [=ordered set=] here... 

```suggestion
          <td>A [=list=] containing the maximum logical value for each axis.
```

Same for the rest that use "array"...

index.html
```diff
@@ -170,6 +170,43 @@ <h2>
...
+          <td><dfn>[[\exposed]]</dfn>
+          <td>`false`
+          <td>A flag indicating that the {{Gamepad}} object has been exposed to script.
+        <tr>
+          <td><dfn>[[\axisMapping]]</dfn>
+          <td>`{}`
+          <td>Mapping from unmapped axis index to an index in the {{Gamepad/axes}} array
+        <tr>
+          <td><dfn>[[\axisMinimums]]</dfn>
+          <td>`null`
```
marcoscaceres: Maybe better to just init to an empty list... then we don't need to check if it's null VS if it's empty.

index.html
```diff
@@ -258,6 +295,81 @@ <h2>
...
+      </p>
+      <ol>
+        <li>Let |gamepad:Gamepad| be the {{Gamepad}} instance in |navigator:Navigator|.{{[[gamepads]]}} representing the gamepad.
+        <li>Let |axisValues:sequence&lt;unsigned long&gt;| be the gamepad's <a>ordered array of logical axis input values</a>.
+        <li>Let |buttonValues:sequence&lt;unsigned long&gt;| be the gamepad's <a>ordered array of logical button input values</a>.
+        <li>Queue a task (where?) to <a>update gamepad state</a> for |gamepad| with |axisValues| and |buttonValues|.
+      </ol>
+      To <dfn>update gamepad state</dfn> for |gamepad| with an <a>ordered array of logical axis input values</a> |axisValues:sequence&lt;unsigned long&gt;| and an <a>ordered array of logical button input values</a> |buttonValues:sequence&lt;unsigned long&gt;|, run the following steps:
+      <ol>
+        <li>Let |now:DOMHighResTimeStamp| be a {{DOMHighResTimeStamp}} representing the current time.
```
```suggestion
        <li>Let |now| be the [=current high resolution time=].
```

index.html
```diff
@@ -258,6 +295,81 @@ <h2>
...
+        <li>Let |gamepadsStart:DOMHighResTimeStamp| be the |navigationStart:DOMHighResTimeStamp| attribute of the {{PerformanceTiming}} interface.
```
marcoscaceres: Hmm... this bit is not right for a couple of reasons:  

 1.  `PerformanceTiming` is just an interface (not an instance)
 2. `PerformanceTiming` was not initiated to anything. 
 3. `navigationStart` is "obsolete".

If we just need "when the global object was created", then:  
  
```suggestion
        <li>Let |gamepadsStart| be the [=time origin=] of the [=global object=].
```

Also, we probably don't need an instance of `DOMHighResTimeStamp` above.

index.html
```diff
@@ -258,6 +295,81 @@ <h2>
...
+        <li>Set |gamepad|.{{Gamepad/timestamp}} to |now| − |gamepadsStart|.
```
```suggestion
        <li>Set |gamepad|.{{Gamepad/timestamp}} to |now| − |gamepadsStart|.
        <p class="warning">User agents SHOULD set a minimum resolution of |gamepad|’s {{Gampad/timestamp}} attribute to 5 microseconds, following [[HR-TIME]]'s [=clock resolution=] recommendation.
        </p>
```

index.html
```diff
@@ -258,6 +295,81 @@ <h2>
...
+        <li>Let |axisCount:long| be the length of |axisValues|.
+        <li>Initialize |rawAxisIndex:long| to be 0.
+        <li>While |rawAxisIndex| < |axisCount|:
```
```suggestion
        <li>While |rawAxisIndex| is less than |axisCount|:
```
😇

index.html
```diff
@@ -258,6 +295,81 @@ <h2>
...
+          <ol>
+            <li>Let |mappedIndex:long| be the value for key |rawAxisIndex| in |gamepad|.{{[[axisMapping]]}}.
```
marcoscaceres: Because we are using an [=ordered map=], Infra let's us do: 

```suggestion
            <li>Let |mapping:ordered map| be |gamepad|.{{[[axisMapping]]}}.
            <li>Let |mappedIndex:long| be |mapping|["rawAxisIndex"].
```
And so on...

index.html
```diff
@@ -258,6 +295,81 @@ <h2>
...
+      <p>
+        When the system receives new input values from a gamepad, run the following steps:
+      </p>
+      <ol>
+        <li>Let |gamepad:Gamepad| be the {{Gamepad}} instance in |navigator:Navigator|.{{[[gamepads]]}} representing the gamepad.
+        <li>Let |axisValues:sequence&lt;unsigned long&gt;| be the gamepad's <a>ordered array of logical axis input values</a>.
+        <li>Let |buttonValues:sequence&lt;unsigned long&gt;| be the gamepad's <a>ordered array of logical button input values</a>.
+        <li>Queue a task (where?) to <a>update gamepad state</a> for |gamepad| with |axisValues| and |buttonValues|.
+      </ol>
+      To <dfn>update gamepad state</dfn> for |gamepad| with an <a>ordered array of logical axis input values</a> |axisValues:sequence&lt;unsigned long&gt;| and an <a>ordered array of logical button input values</a> |buttonValues:sequence&lt;unsigned long&gt;|, run the following steps:
```
marcoscaceres: I think we might want a generalized initializer algorithm for creating Gamepad instances. Check out how I did it in Geo:

https://www.w3.org/TR/geolocation/#constructing-a-geolocationposition

index.html
```diff
@@ -258,6 +295,81 @@ <h2>
...
+            <li>Let |logicalValue:unsigned long| be the value at index |rawAxisIndex| in |axisValues|.
+            <li>Let |logicalMinimum:unsigned long| be the value for key |rawAxisIndex| in |gamepad|.{{[[axisMinimums]]}}.
+            <li>Let |logicalMaximum:unsigned long| be the value for key |rawAxisIndex| in |gamepad|.{{[[axisMaximums]]}}.
+            <li>Let |normalizedValue:double| be 2 (|logicalValue| − |logicalMinimum|) / (|logicalMaximum| − |logicalMinimum|) − 1.
+            <li>Set the value at index |axisIndex| of |gamepad|.{{Gamepad/axes}} to |normalizedValue|.
+            <li>Increment |rawAxisIndex|.
+          </ol>
+        <li>Let |buttonCount:long| be the length of |buttonValues|.
+        <li>Initialize |rawButtonIndex:long| to be 0.
+        <li>While |rawButtonIndex| < |buttonCount|:
```
```suggestion
        <li>While |rawButtonIndex| is less than |buttonCount|:
```

index.html
```diff
@@ -258,6 +295,81 @@ <h2>
...
+          <ol>
+            <li>Let |mappedIndex:long| be the value for key |rawButtonIndex| in |gamepad|.{{[[buttonMapping]]}}.
```
marcoscaceres: Nit: Might be nice to pull these steps and the ones above into a two little algorithms. Pass in |gamepad| and it computes and stores the |normalizedValue|.

index.html
```diff
@@ -258,6 +295,81 @@ <h2>
...
+      </p>
+      <p>
+        The <a>user agent</a> MUST provide methods for retrieving <dfn data-lt="ordered array of minimum logical values">ordered arrays of minimum logical values</dfn> and <dfn data-lt="ordered array of maximum logical values">maximum logical values</dfn> for each gamepad input.
+        The ordering of these arrays MUST match the ordering of the <a>ordered array of axis inputs</a> and <a>ordered array of button inputs</a>.
+      </p>
+      <p>
+        The <a>user agent</a> MUST provide methods for retrieving <dfn data-lt="ordered array of logical axis input values">ordered arrays of logical axis input values</dfn> and <dfn data-lt="ordered array of logical button input values">button input values</dfn>.
+        These arrays MUST contain the most recent values received from the gamepad for each button and axis input.
+        The ordering of these arrays MUST match the ordering of the <a>ordered array of axis inputs</a> and <a>ordered array of button inputs</a>.
+      </p>
```
marcoscaceres: I'd suggest we delete these and just use Infra types. Also, it will avoid descriptors for things that happen algorithmically anyway.

index.html
```diff
@@ -358,68 +470,84 @@ <h2>
     </section>
     <section data-dfn-for="Navigator">
       <h2>
-        Navigator Interface extension
+        Extensions to the {{Navigator}} interface
```
marcoscaceres: Just "code" (backticks) for headings is ok... the Navigator object will be hyperlinked in the IDL:
 
```suggestion
        Extensions to the `Navigator` interface
```

index.html
marcoscaceres: Just a note, only add a `data-lt=` to `dfn`s if you are creating an alias.

Like 
```
<dfn data-lt="fruit of the gods">bananas</dfn>  where the [=fruit of the gods=], and yet...
```

index.html
```diff
@@ -358,68 +470,84 @@ <h2>
...
-          </p>
-          <p>
-            As an example, if there is one connected gamepad with an index of
-            1, then the following code snippet describes the expected behavior:
-          </p>
-          <pre class="example highlight">
+      <p>
+        Instances of {{Navigator}} are created with the internal slots described in the following table:
+      </p>
+      <table class="simple" dfn-for=Navigator dfn-type=attribute>
```
```suggestion
      <table class="simple" dfn-for=Navigator>
```

index.html
```diff
@@ -358,68 +470,84 @@ <h2>
...
+        <tr>
+          <th>Internal slot
+          <th>Initial value
+          <th>Description (non-normative)
+        <tr>
+          <td><dfn>[[\hasGamepadGesture]]</dfn>
```
marcoscaceres: These should be: 

```suggestion
          <td><dfn data-dfn-for="Navigator">[[\hasGamepadGesture]]</dfn>
```

And then... to link... a little ugly... I know: 

```
{{Navigator/[[hasGamepadGesture]]}}
```

marcoscaceres (commented):
Great stuff @nondebug! I didn't manage to get all the way through it yet, but here is the initial set of comments.

index.html
nondebug: Sounds good for these internal arrays. I've left the Navigator.gamepads list as an empty array [] since we need to expose that array to script for getGamepads. Is there some incantation for converting from infra types to Javascript types?

index.html
```diff
@@ -258,6 +278,80 @@ <h2>
...
+      </p>
+      <ol>
+        <li>Let |now| be the [=current high resolution time=].
+        <li>Set |gamepad|.{{Gamepad/timestamp}} to |now|.
+        <li>Run the steps for <a>map and normalize axes</a> with |gamepad|.
+        <li>Run the steps for <a>map and normalize buttons</a> with |gamepad|.
+        <li>If |navigator|.{{Navigator/[[hasGamepadGesture]]}} is `false` and |gamepad| <a>contains a gamepad user gesture</a>:
+          <ol>
+            <li>Set |navigator|.{{Navigator/[[hasGamepadGesture]]}} to `true`.
+            <li>For each |connectedGamepad:Gamepad| of |navigator|.{{Navigator/[[gamepads]]}}:
```
marcoscaceres: Got motivated from our last tele conf and added nullable type support 😇

```suggestion
            <li>For each |connectedGamepad:Gamepad?| of |navigator|.{{Navigator/[[gamepads]]}}:
```

index.html
```diff
@@ -258,6 +278,80 @@ <h2>
...
+              <ol>
+                <li>If |connectedGamepad| is not `null`:
+                  <ol>
+                    <li>Set |connectedGamepad|.{{Gamepad/timestamp}} to |now|.
+                    <li>[[=Fire an event=]] named `"gamepadconnected"` at |window:Window| using {{GamepadEvent}} with its {{GamepadEvent/gamepad}} attribute initialized to |connectedGamepad|.
```
```suggestion
                    <li>[=Fire an event=] named `"gamepadconnected"` at |window:Window| using {{GamepadEvent}} with its {{GamepadEvent/gamepad}} attribute initialized to |connectedGamepad|.
```

index.html
nondebug: I added the warning to the timestamp attribute definition

index.html
nondebug: I've replaced the "value for key..." language with ordered map syntax but decided not to pull out |mapping|, maybe I should add it for readability? WDYT?

index.html
nondebug: Added.

In another review it was pointed out that readonly attributes can't be updated after initialization (because they're read-only) so I've added internal slots for any attributes that may change after initialization and defined getter algorithms that refer to the internal slots.

index.html
nondebug: I've removed these definitions and rewritten the relevant part of the algorithm, please take a look

index.html
nondebug: Thanks!

index.html
nondebug: Fixed

index.html
```diff
@@ -195,69 +335,537 @@ <h2>
...
+                <ol>
+                  <li>If |connectedGamepad| is not equal to `null`:
+                    <ol>
+                      <li>Set |connectedGamepad|.{{Gamepad/[[exposed]]}} to
+                      `true`.
+                      </li>
+                      <li>Set |connectedGamepad|.{{Gamepad/[[timestamp]]}} to
+                      |now|.
+                      </li>
+                      <li>If |gamepad|'s [=relevant global object=] is a
```
marcoscaceres: I think here we should be able to just queue on gamepad task source instead of going through the global one.

index.html
```diff
@@ -195,69 +335,537 @@ <h2>
...
+                      {{Window}} object, [=queue a global task=] on |gamepad|'s
```
marcoscaceres: Here, we might want to also do the fully active check (as we are checking the {{Window}} is still there, per #149).... we can come back to this tho.

index.html
```diff
@@ -195,69 +335,537 @@ <h2>
...
+              </li>
+            </ol>
+          </li>
+        </ol>
+        <p>
+          To <dfn>map and normalize axes</dfn> for |gamepad:Gamepad|, run the
+          following steps:
+        </p>
+        <ol>
+          <li>Let |axisValues:list| be a [=list=] of `unsigned long` values
```
```suggestion
          <li>Let |axisValues:list| be a [=list=] of {{unsigned long}} values
```

index.html
```diff
@@ -195,69 +335,537 @@ <h2>
...
+              </li>
+            </ol>
+          </li>
+        </ol>
+        <p>
+          To <dfn>map and normalize buttons</dfn> for |gamepad:Gamepad|, run
+          the following steps:
+        </p>
+        <ol>
+          <li>Let |buttonValues:list| be a [=list=] of `unsigned long` values
```
```suggestion
          <li>Let |buttonValues:list| be a [=list=] of {{unsigned long}} values
```

index.html
```diff
@@ -195,69 +335,537 @@ <h2>
...
+        <p>
+          To <dfn data-lt="initializing axes">initialize axes</dfn> for
+          |gamepad:Gamepad|, run the following steps:
+        </p>
+        <ol>
+          <li>Let |inputCount:long| be the number of axis inputs exposed by the
+          device represented by |gamepad|.
+          </li>
+          <li>Set |gamepad|.{{Gamepad/[[axisMinimums]]}} to a [=list=] of
+          `unsigned long` values with [=list/size=] equal to |inputCount|
```
```suggestion
          {{unsigned long}} values with [=list/size=] equal to |inputCount|
```

Plus a few other ones...

index.html
```diff
@@ -195,69 +335,537 @@ <h2>
...
+        <p>
+          A [=user agent=] MUST dispatch this event type to indicate the user
+          has connected a gamepad. If a gamepad was already connected when the
+          page was loaded, the {{gamepadconnected}} event SHOULD be dispatched
+          when the user presses a button or moves an axis.
+        </p>
+        <p>
+          A [=user agent=] MUST NOT dispatch this event type if the
+          [=environment settings object=] is a [=non-secure context=].
+        </p>
```
marcoscaceres: We don't need this stuff anymore 🎉... it's already above (including the secure context stuff... the API is not exposed in insecure contexts, so that can't ever happen). 

```suggestion
```

index.html
```diff
@@ -275,43 +883,121 @@ <h2>
           readonly attribute double value;
         };
       </pre>
+      <p>
+        Instances of {{GamepadButton}} are created with the internal slots
+        described in the following table:
+      </p>
+      <table class="simple" dfn-for="GamepadButton">
```
```suggestion
      <table class="simple">
```

index.html
```diff
@@ -722,47 +1502,116 @@ <h2>
...
       </h3>
+      <p>
+        When a gamepad becomes unavailable on the system, run the following
+        steps:
+      </p>
+      <ol>
+        <li>If the [=current settings object=]'s [=environment settings object
+        / responsible document=] is not [=allowed to use=] the `"gamepad"`
+        permission, then abort these steps.
+        </li>
```
marcoscaceres: I don't think we need the check here? 

```suggestion
```

index.html
```diff
@@ -722,47 +1502,116 @@ <h2>
...
+        <li>Let |gamepad:Gamepad| be the {{Gamepad}} representing the
+        unavailable device.
+        </li>
+        <li>Set |gamepad|.{{Gamepad/[[connected]]}} to `false`.
```
marcoscaceres: I think this should be done in the steps of the task (right before the event is fired), no? Otherwise, it would update somewhat randomly.

index.html
```diff
@@ -722,47 +1502,116 @@ <h2>
...
+      </ol>
       <p>
         User agents implementing this specification must provide a new DOM
-        event, named <code>gamepaddisconnected</code>. The corresponding event
-        MUST be of type <code>GamepadEvent</code> and, if <a>allowed to use</a>
-        the "`gamepad`" permission, MUST fire on the <code>window</code>
-        object. Registration for and firing of the
-        <code>gamepaddisconnected</code> event MUST follow the usual behavior
-        of DOM Events. [[DOM]]
+        event, named {{gamepaddisconnected}}. The corresponding event MUST be
```
marcoscaceres, line 1584: I think they "allowed to use" checks should only happen when the developer tries to use the API. 

We can maybe have a global check that if the permission to use the API gets disabled, then just bring down everything... but that's probably overkill.

index.html
```diff
@@ -722,47 +1502,116 @@ <h2>
...
+        of type {{GamepadEvent}} and, if [=allowed to use=] the `"gamepad"`
+        permission, MUST fire on the {{Window}} object. Registration for and
+        firing of the {{gamepaddisconnected}} event MUST follow the usual
+        behavior of DOM Events. [[DOM]]
```
marcoscaceres: Don't need this :) We are doing the right thing... 

```suggestion
        permission, MUST fire on the {{Window}} object.
```

index.html
```diff
@@ -722,47 +1502,116 @@ <h2>
...
+        When a gamepad is disconnected from the [=user agent=], if the [=user
+        agent=] has previously dispatched a {{gamepadconnected}} event for that
+        gamepad to a window, a {{gamepaddisconnected}} event MUST be dispatched
+        to that same window.
       </p>
       <p>
-        A <a>user agent</a> MUST NOT dispatch this event type if the
-        <a>environment settings object</a> is a <a>non-secure context</a>.
+        A [=user agent=] MUST NOT dispatch this event type if the [=environment
+        settings object=] is a [=non-secure context=].
```
marcoscaceres: Don't need this... can't register events in insecure contexts, so they will never fire in these contexts... 

```suggestion
```

marcoscaceres (approved):
Wowser! This is so much better! Nice one @nondebug! 

I'd be inclined for us to take this as is (with the suggestions) and we should just build on it.

index.html

index.html
nondebug: Oops, copy/paste error... removed

index.html
nondebug: Removed.

index.html
nondebug: Removed

index.html
nondebug: Removed

index.html
nondebug: Removed

index.html
nondebug: I've added a fully active check so we won't try to fire events for non-fully-active Documents.

I removed the Window check since I think we'll implicitly make that check when trying to get the associated (or responsible) Document. If the global object isn't a Window then it won't have a Document. Is that correct?

index.html
nondebug: Agreed. We should update the gamepad state synchronously with firing the event to make sure the state changes are handled sequentially and predictably.

I think this order makes the most sense for gamepadconnected:

1. Create gamepad with gamepad.connected initialized to true
2. Add gamepad to navigator.[[gamepads]]
3. Dispatch gamepadconnected

And for gamepaddisconnected:

1. Set gamepad.connected to false
2. Dispatch gamepaddisconnected
3. Remove gamepad from navigator.[[gamepads]]

Updating in this order ensures the gamepad.connected state matches the stated reflected by the gamepadconnected or gamepaddisconnected event, and also ensures the item in the getGamepads() array at index gamepad.index is the same instance as the gamepad instance in the event.

index.html
nondebug: I've removed the "allowed to use" check here but there's still a reference in the gamepadconnected algorithm.

marcoscaceres:
https://www.w3.org/TR/2021/WD-gamepad-20210709/ 🎉

PR #163, opened 2022-06-06
extensions: Add vibrationActuator and dual-rumble effect

nondebug:
Extends the Gamepad API to support common vibration effects.

Implementation commitment:

 * [ ] WebKit (https://bugs.webkit.org/show_bug.cgi?id=248989)

### DIFF
--- a/extensions.html
+++ b/extensions.html
@@ -95,6 +95,8 @@
           // document unless you know what you're doing. If in doubt ask your friendly neighbourhood
           // Team Contact.
           wgPatentURI:  "https://www.w3.org/2004/01/pp-impl/83482/status",
+
+          xref: ["DOM", "HTML", "INFRA", "gamepad"],
       };
     </script>
     <style>
@@ -182,19 +184,142 @@
         <dfn>GamepadHapticActuator</dfn> Interface
       </h2>
       <p>
-        Each <code>GamepadHapticActuator</code> corresponds to a motor or other
-        actuator that can apply a force for the purposes of haptic feedback.
+        A {{GamepadHapticActuator}} corresponds to a configuration of motors or
+        other actuators that can apply a force for the purposes of haptic
+        feedback.
       </p>
       <pre class="idl">
         [Exposed=Window]
         interface GamepadHapticActuator {
           readonly attribute GamepadHapticActuatorType type;
+          boolean canPlayEffectType(GamepadHapticEffectType type);
+          Promise&lt;GamepadHapticsResult&gt; playEffect(
+              GamepadHapticEffectType type,
+              optional GamepadEffectParameters params = {});
           Promise&lt;boolean&gt; pulse(double value, double duration);
+          Promise&lt;GamepadHapticsResult&gt; reset();
         };
       </pre>
+      <p>
+        Instances of {{GamepadHapticActuator}} are created with the internal
+        slots described in the following table:
+      </p>
+      <table class="simple">
+        <tr>
+          <th>
+            Internal slot
+          </th>
+          <th>
+            Initial value
+          </th>
+          <th>
+            Description (non-normative)
+          </th>
+        </tr>
+        <tr>
+          <td>
+            <dfn data-dfn-for=
+            "GamepadHapticActuator">[[\playingEffectPromise]]</dfn>
+          </td>
+          <td>
+            `undefined`
+          </td>
+          <td>
+            The {{Promise}} for the playing effect, or `undefined` if no effect
+            is playing.
+          </td>
+        </tr>
+      </table>
+      <dl>
+        <dt>
+          <dfn>type</dfn> attribute
+        </dt>
+        <dd>
+          <p>
+            A {{GamepadHapticActuatorType}} value representing the actuator
+            type.
+          </p>
+        </dd>
+      </dl>
       <dl data-dfn-for="GamepadHapticActuator">
         <dt>
-          <dfn>pulse()</dfn>
+          <dfn>canPlayEffectType()</dfn> method
+        </dt>
+        <dd>
+          <p>
+            The {{GamepadHapticActuator/canPlayEffectType()}} method steps are:
+          </p>
+          <ol>
+            <li>If this actuator can [=play effects with type=]
+            |type:GamepadHapticEffectType|, return `true`.
+            </li>
+            <li>Otherwise, return `false`.
+            </li>
+          </ol>
+        </dd>
+        <dt>
+          <dfn>playEffect()</dfn> method
+        </dt>
+        <dd>
+          <p>
+            The {{GamepadHapticActuator/playEffect()}} method steps are:
+          </p>
+          <ol>
+            <li>If |params:GamepadEffectParameters| does not describe a [=valid
+            effect=] of type |type:GamepadHapticEffectType|, return [=a promise
+            resolved with=] {{GamepadHapticsResult/"invalid-parameter"}}.
+            </li>
+            <li>Let |document| be the [=current settings object=]'s [=relevant
+            global object=]'s [=associated `Document`=].
+            </li>
+            <li>If |document| is `null` or |document| is not [=Document/fully
+            active=] or |document|'s [=visibility state=] is `"hidden"`, return
+            [=a promise resolved with=] {{GamepadHapticsResult/"preempted"}}.
+            </li>
+            <li>If [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}}
+            is not `undefined`:
+              <ol>
+                <li>[=Queue a global task=] on the [=relevant global object=]
+                of [=this=] using the [=gamepad task source=] to [=resolve=]
+                [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}}
+                with {{GamepadHapticsResult/"preempted"}}.
+                </li>
+                <li>Set
+                [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}} to
+                `undefined`.
+                </li>
+              </ol>
+            </li>
+            <li>If this actuator cannot [=play effects with type=] |type|,
+            return [=a promise resolved with=]
+            {{GamepadHapticsResult/"not-supported"}}.
+            </li>
+            <li>Let {{GamepadHapticActuator/[[playingEffectPromise]]}} be [=a
+            new promise=].
+            </li>
+            <li>[=Issue a haptic effect=] to the actuator with |type| and
+            |params|.
+            </li>
+            <li>Return {{GamepadHapticActuator/[[playingEffectPromise]]}}.
+            </li>
+          </ol>
+          <p>
+            When the effect completes, if it was not preempted, run these
+            steps:
+          </p>
+          <ol>
+            <li>[=Queue a global task=] on the [=relevant global object=] of
+            [=this=] using the [=gamepad task source=] to [=resolve=]
+            [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}} with
+            {{GamepadHapticsResult/"complete"}}.
+            </li>
+            <li>Set [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}}
+            to `undefined`.
+            </li>
+          </ol>
+        </dd>
+        <dt>
+          <dfn>pulse()</dfn> method
         </dt>
         <dd>
           <p>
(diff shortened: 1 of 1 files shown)


nondebug:
This is intended to replace https://github.com/w3c/gamepad/pull/68 which can't be easily rebased on the current version of the spec.

marcoscaceres:
Just to make it easier to review, it might be good to do a tidy of the underlying document and then send these the changes on top.

marcoscaceres:
Sent https://github.com/w3c/gamepad/pull/164

extensions.html
```diff
@@ -182,19 +184,142 @@ <h2>
...
+        </dd>
+        <dt>
+          <dfn>playEffect()</dfn> method
+        </dt>
+        <dd>
+          <p>
+            The {{GamepadHapticActuator/playEffect()}} method steps are:
+          </p>
+          <ol>
+            <li>If |params:GamepadEffectParameters| does not describe a [=valid
```
marcoscaceres, line 268: Can you help me understand this a bit.... wouldn't rejecting with a `TypeError` be better here? it seems a bit odd to resolve successfully for what is actually an error.

extensions.html
```diff
@@ -182,19 +184,142 @@ <h2>
...
+            effect=] of type |type:GamepadHapticEffectType|, return [=a promise
+            resolved with=] {{GamepadHapticsResult/"invalid-parameter"}}.
+            </li>
+            <li>Let |document| be the [=current settings object=]'s [=relevant
+            global object=]'s [=associated `Document`=].
+            </li>
+            <li>If |document| is `null` or |document| is not [=Document/fully
```
marcoscaceres, line 275: Same here...

extensions.html
```diff
@@ -182,19 +184,142 @@ <h2>
...
+            active=] or |document|'s [=visibility state=] is `"hidden"`, return
+            [=a promise resolved with=] {{GamepadHapticsResult/"preempted"}}.
+            </li>
+            <li>If [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}}
```
marcoscaceres, line 279: As above, why not just return a new rejected promise here?

Also, this is seems to be racy. This is called twice, `[[playingEffectPromise]]` could resolve before the second call.

extensions.html
```diff
@@ -182,19 +184,142 @@ <h2>
...
+                [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}}
+                with {{GamepadHapticsResult/"preempted"}}.
+                </li>
+                <li>Set
+                [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}} to
+                `undefined`.
+                </li>
+              </ol>
+            </li>
+            <li>If this actuator cannot [=play effects with type=] |type|,
```
marcoscaceres, line 294: As above... this feels like it should be a `"NotSupportedError"` DOMException.

extensions.html
```diff
@@ -182,19 +184,142 @@ <h2>
...
+            return [=a promise resolved with=]
+            {{GamepadHapticsResult/"not-supported"}}.
+            </li>
+            <li>Let {{GamepadHapticActuator/[[playingEffectPromise]]}} be [=a
```
marcoscaceres, line 297: I think here you want "Do the following steps in parallel:" ... request actuator do stuff in sub steps... return the promise in main steps.

extensions.html
```diff
@@ -182,19 +184,142 @@ <h2>
...
+            new promise=].
+            </li>
+            <li>[=Issue a haptic effect=] to the actuator with |type| and
+            |params|.
+            </li>
+            <li>Return {{GamepadHapticActuator/[[playingEffectPromise]]}}.
+            </li>
+          </ol>
+          <p>
```
marcoscaceres: Right, then this can all move into the "in parallel" sub-steps.

extensions.html
```diff
@@ -182,19 +184,142 @@ <h2>
...
+            When the effect completes, if it was not preempted, run these
+            steps:
+          </p>
+          <ol>
+            <li>[=Queue a global task=] on the [=relevant global object=] of
+            [=this=] using the [=gamepad task source=] to [=resolve=]
+            [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}} with
+            {{GamepadHapticsResult/"complete"}}.
+            </li>
+            <li>Set [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}}
```
marcoscaceres: Clearing [[playingEffectPromise]] should be done as part of the task. Otherwise, it reads like [[playingEffectPromise]] gets trashed before the task runs.

extensions.html
```diff
@@ -209,33 +334,373 @@ <h2>
...
-      </p>
-      <pre class="idl">
-        enum GamepadHapticActuatorType {
-          "vibration"
-        };
-      </pre>
-      <dl data-dfn-for="GamepadHapticActuatorType">
         <dt>
-          <dfn>vibration</dfn>
+          <dfn data-lt="reset()">reset</dfn>
```
```suggestion
          <dfn>reset()</dfn>
```

extensions.html
nondebug: This PR is intended to describe the current implementation in Chromium which resolves with "invalid-parameter":

https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/gamepad/gamepad_haptic_actuator.cc;l=95;drc=db9ae7941adc1d95c943accce9e0151d265fd640

I agree it would make more sense to reject with `TypeError`. I think we can change this without breaking applications since there aren't likely to be any applications that depend on the current behavior.

extensions.html
nondebug: Which error would you use?

I feel a preempted effect is not an error because it's expected you'll sometimes interrupt an effect to cut it short or replace it with a new effect. I think visibility changes should be handled as if the effect were preempted by a call to `reset()`.

extensions.html
nondebug: `NotSupportedError` seems reasonable here.

extensions.html
nondebug: Please take another look

extensions.html
nondebug: Done

extensions.html
marcoscaceres: Understood. The case here feels different because this step is happening as a result of an explicit call to `.playEffect()`. The change in document visibility resolves the `[[playingEffectPromise]]` with "preempted", but that's different from this call. 

Note: in addition the visibility change handler below, you can add some steps to watch for the document becoming not fully active. The visibility change and the document not being fully active can then have the same behavior.

marcoscaceres (commented):
As a general comment, might be interesting to consider adding `AbortController` support to `playEffect()`. For instance, the user hits the menu button, then it would be nice to just abort the current effect.

extensions.html
marcoscaceres: Also, with `AbortController` support, then `.reset()` can effectively go away. It's all controller from a single place.

extensions.html
```diff
@@ -182,19 +184,146 @@ <h2>
...
       </p>
       <pre class="idl">
         [Exposed=Window]
         interface GamepadHapticActuator {
           readonly attribute GamepadHapticActuatorType type;
+          boolean canPlayEffectType(GamepadHapticEffectType type);
+          Promise&lt;GamepadHapticsResult&gt; playEffect(
+              GamepadHapticEffectType type,
+              optional GamepadEffectParameters params = {});
           Promise&lt;boolean&gt; pulse(double value, double duration);
```
marcoscaceres, line 199: isn't "pulse" also an effect? Maybe it could be `playEffect("pulse")`... otherwise, you end up with situations like:

```
.pulse(a,b);
.playEffect(obj);
```

Which would conflict with each other. Having a single `.playEffect()` call would avoid that. 

```
.playEffect("dual-rumble").then(()=> .playEffect("pulse"))
```

extensions.html
```diff
@@ -182,19 +184,146 @@ <h2>
...
-        actuator that can apply a force for the purposes of haptic feedback.
+        A {{GamepadHapticActuator}} corresponds to a configuration of motors or
+        other actuators that can apply a force for the purposes of haptic
+        feedback.
       </p>
       <pre class="idl">
         [Exposed=Window]
         interface GamepadHapticActuator {
           readonly attribute GamepadHapticActuatorType type;
+          boolean canPlayEffectType(GamepadHapticEffectType type);
```
marcoscaceres, line 195: Maybe just? 

```suggestion
          boolean canPlay(GamepadHapticEffectType type);
```

extensions.html
```diff
@@ -182,19 +184,146 @@ <h2>
...
+                </li>
+              </ol>
+            </li>
+            <li>If this actuator cannot [=play effects with type=] |type|,
+            return [=a promise rejected with=] reason {{NotSupportedError}}.
+            </li>
+            <li>Let {{GamepadHapticActuator/[[playingEffectPromise]]}} be [=a
+            new promise=].
+            </li>
+            <li>Do the following steps in parallel:
```
marcoscaceres, line 300:
```suggestion
            <li>Do the following steps [=in parallel=]:
```

extensions.html
nondebug: I'm fine with removing pulse() since I think there are no actual implementations. Chromium doesn't implement it, Firefox has IDL but it's not hooked up on any platforms.

I'm not sure we even need a "pulse" effect. In practice, far more gamepads use two-channel "dual-rumble" haptics than single-channel "pulse" haptics. Reading the [Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=680289) it's not clear to me how we arrived at the single-channel version currently in extensions.html.

I'd lean toward removing pulse() and having applications send "dual-rumble" effects with equal strongMagnitude and weakMagnitude values since that's what the user agent would likely do anyway.

extensions.html
marcoscaceres: Actually, I think `.supports()` might be better. Consider, just because gamepad supports an effect type, doesn't mean it can play it. 

```suggestion
          boolean supports(GamepadHapticEffectType type);
```

marcoscaceres: To future-proof this, this should take a `DOMString`, not a enum value. Otherwise, this will throw when called with unknown value in the future (affecting legacy engines).

The logic here being that you should be able to do:

```JS
const myGamesEffects = ["rumble", "bumble", "trigger-me-fumble", "dual-rumble", "pulse"]; 
// This shouldn't throw... "trigger-me-fumble" is supported in FutureBrowser 2000!
const supportedEffects = myGamesEffects.filter(effect => gamepad.vibrationActuator.canPlay(effect));
```

Alternatively, why not just ask the `vibrationActuator` to return what it supports via a frozen array? 
Then there is no need to keep asking the gamepad what it can do. 

```JS
gamepad.vibrationActuator.effects.includes("dual-rumble")
```

extensions.html
marcoscaceres: Agree. It would be nice to have a single way to activate the actuator.

extensions.html
```diff
@@ -182,19 +184,146 @@ <h2>
...
+            Initial value
+          </th>
+          <th>
+            Description (non-normative)
+          </th>
+        </tr>
+        <tr>
+          <td>
+            <dfn data-dfn-for=
+            "GamepadHapticActuator">[[\playingEffectPromise]]</dfn>
```
marcoscaceres, line 222: This guard might be better as just a boolean/flag to control the state machine. This is because the actual promise to play something can only be returned once.

However, I'm worried about something... 

Consider the case where you want to do the following:

 1. Lightly "rumble" the right-side of the controller. 
 2. Simultaneously, "rumble" one of the triggers.
 
In theory, it should be possible to activate multiple actuators at the same time, no? 

The current design would prevent this from working:

```JS
const rumble = gamepad.vibrationActuator.play("double-rumble", {intensity: 0.2, duration: 2000});
// reject :( 
const trigger = gamepad.vibrationActuator.play("trigger-rumble", {intensity: 2.0, duration: 2000});  
```

So, I think the guard is actually a Map `map<GamepadHapticActuatorType, boolean>`.  Something like: 

```
GamepadHapticActuator ["dual-rumble"] -> isPlaying ?
GamepadHapticActuator["left-trigger"] -> isPlaying ?
GamepadHapticActuator["right-trigger"] -> isPlaying ?
```

extensions.html
```diff
@@ -209,33 +338,356 @@ <h2>
...
+                </li>
+                <li>Set
+                [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}} to
+                `undefined`.
+                </li>
+              </ol>
+            </li>
+            <li>[=Stop haptic effects=] on this actuator.
+            </li>
+            <li>Return [=a promise resolved with=]
```
marcoscaceres, line 373: The promise should probably only resolve once it gets confirmation over IPC that the actuator has actually stopped.

extensions.html
```diff
@@ -209,33 +338,356 @@ <h2>
...
+        <dd>
+          <p>
+            Add new steps to [=a new `Gamepad`=]:
+          </p>
+          <ol>
+            <li>If |gamepad:Gamepad| has a vibration actuator that can [=play
+            effects with type=] {{GamepadHapticEffectType/"dual-rumble"}},
+            initialize |gamepad:Gamepad|'s {{Gamepad/vibrationActuator}}
+            attribute to [=a new `GamepadHapticActuator`=] with
+            {{GamepadHapticActuatorType/"dual-rumble"}}.
```
marcoscaceres, line 390: Hmm.. I'm a bit confused... the `GamepadHapticActuator` seems to be both an actuator "hub" but also a place play the effects. This doesn't feel right. 


Shouldn't it be:

```
Gamepad 
   -> actuator === "all the possible actuators"
        ->  "dual-rumble" === "play effect on the handles"
        ->  "xxxxx-rumble" === "play xxxxx on some part of the controller"
```

What's confusing is the `.type` attribute. What does it mean for `.type` to be "dual-rumble" if I can play "trigger-rumble"?

extensions.html
```diff
@@ -209,33 +338,356 @@ <h2>
...
+                |actuator|.{{GamepadHapticActuator/[[playingEffectPromise]]}}
+                with {{GamepadHapticsResult/"preempted"}}.
+                </li>
+                <li>Set
+                |actuator|.{{GamepadHapticActuator/[[playingEffectPromise]]}}
+                to `undefined`.
+                </li>
+              </ol>
+            </li>
+            <li>[=Stop haptic effects=] on |actuator|.
```
marcoscaceres, line 443: See above... this should probably first stop the effects, get IPC confirmation, and then resolve `playingEffectPromise` with "preempted". Otherwise, this will be racy (stopping the effects might take longer than resolving the promise).

marcoscaceres (changes requested):
Let's put this on the agenda for the next call. I've got questions (concerns?) about `GamepadHapticActuator` itself.

extensions.html
nondebug: If I understand correctly this would be a new (optional) effect param and would let you do something like this:

    const controller = new AbortController();
    const playEffectPromise = gamepad.vibrationActuator.playEffect('dual-rumble', {signal:controller.signal, strongMagnitude:1.0, duration:1000});
    controller.abort(); // instead of gamepad.vibrationActuator.reset()
    const effectResult = await playEffectPromise; // resolves to 'preempted'

That sounds good to me.

extensions.html
nondebug: Either `DOMString` or returning a `FrozenArray` sound good to me. I slightly prefer `gamepad.vibrationActuator.effects` because it makes it possible to automatically discover new effect types.

extensions.html
nondebug: > In theory, it should be possible to activate multiple actuators at the same time, no?

With the current proposal for trigger rumble, it should work to issue a "trigger-rumble" effect that contains the parameters from the "dual-rumble" effect along with the parameters for the trigger effect. Something like:

```
// rumble and vibrate right trigger
const effectParams = {rightTrigger: 1.0, strongMagnitude: 0.2, weakMagnitude: 0.2, duration: 2000};
gamepad.vibrationActuator.play("trigger-rumble", effectParams);
```

The actuators can be controlled independently by preempting the ongoing effect:

```
// stop trigger vibration, keep rumble
effectParams.rightTrigger = 0.0;
gamepad.vibrationActuator.play("trigger-rumble", effectParams);
```

I think the API should not support simultaneous effects on a single GamepadHapticActuator. If there is an ongoing effect when a new effect is played, the ongoing effect is canceled before the new effect begins.

extensions.html
nondebug: I think we (Chromium) should deprecate and remove the actuator's `type` attribute since it's not useful for capability detection. Going forward, all vibration actuators that support "dual-rumble" effects should use type "dual-rumble" even if they support other effects. Once we've provided an alternate means of capability detection (either `canPlay` or `actuator.effects.includes`) then we can remove the `type` attribute.

In general, I don't want compatibility to be limited by the actuator type. It should be possible for user agents to emulate effects even when the actuator doesn't natively support that type of effect. For instance, a Switch Pro controller with an LRA in each handle might be able to play a "dual-rumble" effect meant for ERM actuators if the user agent knows how to simulate a dual-rumble effect with LRAs.

I agree `GamepadHapticActuator` is confusing. In my view it doesn't represent a specific physical actuator on the gamepad, it represents a system of actuators that can work together to create vibration effects. If a `GamepadHapticActuator` supports an effect type, that means its set of actuators includes the appropriate actuators to simulate the effect.

I originally thought that the `type` would help applications understand which effect type is preferred by the actuator. For example, a user agent might allow a "vibration" actuator to play "dual-rumble" effects as a fallback, with some loss in fidelity. The `type` attribute would let applications know to prefer "vibration" effects over "dual-rumble". I've changed my mind and I don't think this is useful anymore.

extensions.html
nondebug: I'm not sure if this is possible. Typically we don't get any feedback from the gamepad when an actuator is started or stopped. The best we can do is confirm that the command was sent to the gamepad.

extensions.html
cabanier: The quest browser implemented pulse and it's widely use by WebXR experiences so it ca't be removed

marcoscaceres (approved):

cabanier:
@marcoscaceres @nondebug I don't see these changes reflected in the spec. How long does it take before they're published?

marcoscaceres:
Oh, they are in the extension spec right now, but we plan to move them over 🔜.

PR #168, opened 2022-07-19
Enhance Gamepad interface description for Touch

### LINKED ISSUE #167: Update subsections "Receiving inputs" and "Constructing a Gamepad" of Gamepad Interface for touch inputs.
Section "Partial Gamepad Interface" is missing subsections  "Receiving inputs" and "Constructing a Gamepad" as it relates to inputs from a touch surface. 
In the main spec the subsections describe the steps followed when "the system receives new button or axis input values".
This spec should provide the same information for when the system receives touch input values.

I have already made the changes and can submit a PR, however to follow the process I believe that this issue should be created first and approved.

@sagoston and @fernando-80 requested I make these changes.

bradleyneedham:
Closes #167 

Implementation commitment:

 * [x] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=1143942#c14)

Standards positions:
* [x] Webkit (https://github.com/WebKit/standards-positions/issues/110)
* [ ] Gecko (https://github.com/mozilla/standards-positions/issues/724)

Resources:
* [x] Explainer (https://github.com/bradleyneedham/gamepad/blob/multitouch-explainer/multitouch_explainer.md)
* [x] Gamepad Tests (https://github.com/bradleyneedham/gamepadtest)
* [x] Tag Review (https://github.com/w3ctag/design-reviews/issues/799)


siusin:
@bradleyneedham could you [associate your GitHub account with your W3C account](https://www.w3.org/users/myprofile/connectedaccounts), please? It will help to resolve the IRP issue. Thank you.

bradleyneedham:
Absolutely would if I had one :-). Does this work better if @sagoston submits the changes?
Signed up for an account. No email to confirm yet. Will update here once resolved.

bradleyneedham:
Ok, done, GitHub account should now be associated with my W3C account.

siusin:
@bradleyneedham 

Excellent, thank you! The editors are expected to review this pull request and merge it when ready.

extensions.html
```diff
@@ -445,16 +507,122 @@ <h2>
...
+              |normalizedX:float|.
+              </li>
+              <li>Set |touchEvent|.{{GamepadTouch/position}}[1] to be
+              |normalizedY:float|.
+              </li>
+              <li>Set |touchEvent|.{{GamepadTouch/surfaceDimensions}}[0] to be
+              device max x-axis coordinate.
+              </li>
+              <li>Set |touchEvent|.{{GamepadTouch/surfaceDimensions}}[1] to be
+              device max y-axis coordinate.<p>(<i>note: surfaceDimensions can be set to `null`</i>)</p>
```
nondebug: Please rewrite these steps with a branch where surfaceDimensions is set to null. We can leave "why is it null?" up to the browser implementer, but if we don't have steps where it can be set to null then it should never be null.

It should be something like:

- If the touch surface exposes maximum surface dimensions in device units:
  - Set touchEvent.surfaceDimensions[0] to ...
  - Set touchEvent.surfaceDimensions[1] to ...
- Otherwise:
  - Set touchEvent.surfaceDimensions to null

extensions.html
```diff
@@ -445,16 +507,122 @@ <h2>
...
+              </li>
+              <li>Add |touchEvent| to |gamepad|.{{Gamepad/[[touchEvents]]}}.
```
nondebug: We're adding to the touchEvents internal slot but never remove from it. This algorithm should start by emptying touchEvents since we plan to replace the whole list each time we update.

extensions.html
```diff
@@ -445,16 +507,122 @@ <h2>
...
+                </ul>
+                </p>
+              </li>
+              <li>Let |touchEvent:GamepadTouch| be a {{GamepadTouch}}.
+              </li>
+              <li>Set |touchEvent|.{{GamepadTouch/touchId}} to be
+              `touchData.touchID`.
+              </li>
+              <li>Set |touchEvent|.{{GamepadTouch/surfaceId}} to be
+              device surfaceId.
```
nondebug: It's unclear where the device surfaceId comes from. I think it would be simpler to specify it as an index into whatever the surface ordering happens to be:

- Let surfaceId be 0
- For each touch surface:
  - For each touch input on that surface:
    - Create and add the touch event (with surfaceId)
  - Increment surfaceId

extensions.html
```diff
@@ -445,16 +507,122 @@ <h2>
...
+                <p>Possible implementation:</p>
+                <ul>
+                  <li>`normalizedY = (2.0 * touchData.y / surfaceDimensionY) - 1`</li>
+                </ul>
+                </p>
+              </li>
+              <li>Let |touchEvent:GamepadTouch| be a {{GamepadTouch}}.
+              </li>
+              <li>Set |touchEvent|.{{GamepadTouch/touchId}} to be
+              `touchData.touchID`.
```
nondebug: We need to specify more of how the touch event's touchID is initialized since it's important for how applications will interact with the API. Let's try to leverage the Touch Events specification which already defines "active touch point":

https://w3c.github.io/touch-events/#dfn-active-touch-point

- If the touch data is part of an existing [=active touch point=] tracked by the user agent:
  - Set touchID to the touchID of the active touch point
  - Otherwise, set touchID to nextTouchID and increment nextTouchID

nextTouchID should be a Gamepad internal slot so it's always unique even when there are multiple touch surfaces on the same gamepad.

extensions.html
```diff
@@ -445,16 +507,122 @@ <h2>
...
-          <code>null</code>.
+          `null`.
         </dd>
         <dt>
           <dfn>touchEvents</dfn>
         </dt>
         <dd>
           A list of touch events generated from all touch surfaces.
-          <code>null</code> if the device does not support touch events.
+          If the device does not support touch events, MUST be set to `null`.
```
nondebug: touchEvents should define a getter algorithm that returns the data from the internal slot.

extensions.html
```diff
@@ -445,16 +507,122 @@ <h2>
...
+        </h3>
+        <p>
+          This section supplements the <a href="https://w3c.github.io/gamepad/#receiving-inputs">
+          Receiving inputs</a> section of the main 
+          <a href="https://w3c.github.io/gamepad/">Gamepad specification</a>.
+        </p>
+        <p>
+          In addtion to the steps defined in the main 
+          <a href="https://w3c.github.io/gamepad/">Gamepad specification</a>.
+          When the system receives new touch input values,
```
nondebug: > When the system receives new touch input values,

This seems ambiguous to me, it's unclear whether "new touch input values" includes updates where there are no touches, for example a DS4 input report that doesn't contain any active touch points. We want this method to run for every DS4 input report, even when there are no active touch points.

Perhaps something like:

"When the user agent modifies the list of [=active touch points=] for a touch surface by adding touch points, removing touch points, or updating the values of existing touch points, update touchEvents by running the following steps:"

extensions.html
```diff
@@ -445,16 +507,122 @@ <h2>
...
+        <p>
+          In addtion to the steps defined in the main 
+          <a href="https://w3c.github.io/gamepad/">Gamepad specification</a>.
+          When constructing <dfn>a new `Gamepad`</dfn> representing a connected gamepad device
+          perform the following steps:
+        </p>
+        <ol>
+          <li>Let |gamepad:Gamepad| be a newly created {{Gamepad}} instance:
+            <ol>
+              <li>Initialize |gamepad|'s {{Gamepad/touchEvents}} to the result of
```
nondebug: I think here we should initialize the touchEvents internal slot to an empty list if the gamepad has any touch surfaces, otherwise null. The [=updating touchEvents=] algorithm will be called automatically whenever the first touch inputs are received so we don't need to call it here.

The actual touchEvents attribute (not the internal slot) should define a getter algorithm that returns the data from the internal slot. That way the attribute is initialized when we initialize the internal slot. We use the internal slot because the attribute is readonly so technically it has no setter but we still need a way to update it.

extensions.html
bradleyneedham: Updated.

extensions.html
bradleyneedham: Added step to empty out any existing events from touchEvents.

extensions.html
bradleyneedham: Added outer loop to iterate surfaces.

extensions.html
bradleyneedham: This may need more discussion. Perhaps there is not a case where surfaceDimensions can be null. For the DS4 touch coordinates are normalized using the device touch point and the surface dimensions. Should the algorithm include receiving input from a device that sends already normalized touch points and does not report surface dimensions?

extensions.html
bradleyneedham: Updated.

extensions.html
bradleyneedham: Added.

extensions.html
bradleyneedham: Updated.

extensions.html
bradleyneedham: @nondebug I manually added the link, it appears that the dfn is not exported (or possibly, probably, I am not using xref correctly). I am also a little concerned that we are making the spec too restrictive to implementers, e.g. I am working on the DS4 GamepadTouch implementation in chromium and find that normalizing the touchId is much easier at the ds4_controller than the device/gamepad (this may be a bad assumption on my part that when you refer to an "internal slot" it would translate to that object). Requiring the nextTouchId in device/gamepad might cause difficulty in implementing other touch controllers. The specification just needs to be clear that implementers support 32 bit value starting at 0 and incrementing with each new touch event not associated with an existing active touch point. How this is accomplished should be open.
I will try and add some "Usage Examples" to better explain how applications will interact with the API.

nondebug: > it appears that the dfn is not exported

I'm not sure why the xref dfn doesn't work. It doesn't show up in https://respec.org/xref/ either. https://respec.org/docs/#xref suggests it needs a change to the touch-events spec to export it. I'm okay with leaving it as-is for now, we can clean this up later when the extension is merged into the main specification.

> normalizing the touchId is much easier at the ds4_controller than the device/gamepad (this may be a bad assumption on my part that when you refer to an "internal slot" it would translate to that object)

Normalizing touchId in Dualshock4Controller won't work because a device::Gamepad isn't 1-to-1 with a blink::Gamepad. Most of the time, "internal slot" should be read as "1-to-1 with blink::Gamepad". The only time we can attach internal slots to device::Gamepad is if all tabs should see the same value. This isn't true for touchId and nextTouchId since they may differ between tabs even for the same gamepad.

Suppose there's a tab accessing a DS4 through Gamepad API and it receives a touch point with ID 0. Then a second tab is opened and begins accessing DS4 through Gamepad API. When the next touch occurs, the first tab should get a touch point with ID 1 and the second tab should get touch ID 0. If we deliver touch ID 1 to both tabs then the second tab can deduce that another tab was accessing Gamepad API.

We can store raw touchIds in the device::Gamepad buffer and normalize them in blink::Gamepad::UpdateFromDeviceState. To normalize the ID, blink::Gamepad will need to maintain nextTouchId along with a mapping from raw->normalized touchIds for all active touch points. The rest of the touch data (surfaceId, position, surfaceDimensions) won't change between tabs so we can store it in the device::Gamepad buffer.

> Requiring the nextTouchId in device/gamepad might cause difficulty in implementing other touch controllers. The specification just needs to be clear that implementers support 32 bit value starting at 0 and incrementing with each new touch event not associated with an existing active touch point. How this is accomplished should be open.

Agreed, the spec should not care how active touch points are identified. We only need to be able to enumerate them and somehow associate current set of active touch points with the set of active touch points from the previous update.

extensions.html
```diff
@@ -445,16 +507,168 @@ <h2>
...
+          When the user agent modifies the list of |active touch points| for a 
+          touch surface by adding touch points, removing touch points, 
+          or updating the values of existing touch points,
+          <dfn data-lt="updating touchEvents">update touchEvents</dfn>
+          by running the following steps:
+        </p>
+        <ol>
+          <li>Let |surfaceId:unsigned long| be 0.
+          </li>
+          <li>Let |touchSurfaces:list| be an [=ordered set=] of touch surfaces
```
nondebug: In this algorithm we describe `touchSurfaces`, `touchInputs`, `touchData` but there's no instructions on how to construct those objects.

We want to emphasize the assumptions we're making about the inputs but we also don't want to specify more than we have to. I think we can modify the algorithm to be more specific about the assumptions we're making while being less specific about how the data is presented. For example, if we specify that a gamepad's touch surfaces need to have some consistent ordering then we can iterate over them without creating lists:

* Let |surfaceId| be 0.
* Repeat these steps for each [=touch surface=] on |gamepad| in [=touch surface enumeration order=]:
  * Let |surfaceDimensions| be `null`.
  * If the [=touch surface=] exposes maximum dimensions in device units:
    * Set |surfaceDimensions| to a [=new=] {{Array}} with length 2.
    * Set |surfaceDimensions|[0] to the maximum X dimension of the current [=touch surface=] in device units.
    * Set |surfaceDimensions|[1] to the maximum Y dimension of the current [=touch surface=] in device units.
  * Repeat these steps for each [=active touch point=] on the current [=touch surface=]:
    * Let |x:float| be the X coordinate for the current [=active touch point=] in device units.
    * Let |y:float| be the Y coordinate for the current [=active touch point=] in device units.
    * Let |position| be a [=new=] {{Array}} with length 2.
    * If |surfaceDimensions| is `null`:
      * Set |position|[0] to ???.
      * Set |position|[1] to ???.
    * Otherwise:
      * Set |position|[0] to a value calculated from |x| and |surfaceDimensions|[0] and normalized to [-1,1].
      * Set |position|[1] to a value calculated from |y| and |surfaceDimensions|[1] and normalized to [-1,1].
    * Let |touchEvent| be a [=new=] {{GamepadTouch}}:
      * Initialize |touchEvent|.{{GamepadTouch/surfaceId}} to |surfaceId|.
      * Initialize |touchEvent|.{{GamepadTouch/position}} to |position|.
      * Initialize |touchEvent|.{{GamepadTouch/surfaceDimensions}} to |surfaceDimensions|.
      * [=list/Append=] |touchEvent| to |gamepad|.{{Gamepad/[[touchEvents]]}}.
  * Increment |surfaceId|.

bradleyneedham: > Normalizing touchId in Dualshock4Controller won't work because a device::Gamepad isn't 1-to-1 with a blink::Gamepad. Most of the time, "internal slot" should be read as "1-to-1 with blink::Gamepad". The only time we can attach internal slots to device::Gamepad is if all tabs should see the same value. This isn't true for touchId and nextTouchId since they may differ between tabs even for the same gamepad.

I went back and double checked and what I have observed is that the lifetime of a blink::Gamepad is very short maybe each render, so will not work for touchId and nextTouchId. This is probably not the forum for this discussion (I need to figure out how to start it over on the chromium site).
This does create an issue with implementing the specification since the lifetime of the device::Gamepad is not correct either as you point out.

> We want to emphasize the assumptions we're making about the inputs but we also don't want to specify more than we have to. I think we can modify the algorithm to be more specific about the assumptions we're making while being less specific about how the data is presented. For example, if we specify that a gamepad's touch surfaces need to have some consistent ordering then we can iterate over them without creating lists:

Agreed, I will work on addressing this.

nondebug: Ah right, blink::Gamepad is short-lived. The specification says it should behave like a live object but the Chromium implementation hasn't been updated to match that behavior yet. (Relevant spec issue: https://github.com/w3c/gamepad/issues/8)

We've worked around this by attaching the internal state to blink::NavigatorGamepad instead of blink::Gamepad and creating a Gamepad::Client interface to provide access to this internal state so it's persisted even when the blink::Gamepad is destroyed and recreated.

> This is probably not the forum for this discussion (I need to figure out how to start it over on the chromium site).

Let's use this crbug to discuss the Chromium implementation: [1143942: Gamepad Multitouch](https://crbug.com/1143942)

bradleyneedham: @nondebug I have restructured the Receiving inputs section and added a Glossary for terms. Several updates to your other comments we made a bit back, should I just resolve those comments? 
There is a suggestion at the top "Still in progress? Convert to draft", does that make since?

extensions.html
```diff
@@ -445,16 +507,166 @@ <h2>
...
+                  </li>
+                  <li>
+                    <p>|position|[1] should be set to the device Y coordinate relative to the device [=touch surface=] and normalized to [-1.0,1.0]
+                      where -1.0 is the topmost coordinate and 1.0 is the bottommost coordinate.
+                    </p>
+                    <p class="note" title="Possible implementation (if surfaceDimensions are available)">
+                      `position[1] = (2.0 * touchData.y / surfaceDimensions[1]) - 1`
+                    </p>
+                  </li>
+                  </li>
```
nondebug: Please remove the extra `</li>`.

extensions.html
```diff
@@ -445,16 +507,166 @@ <h2>
...
+                  <li>Let |touchEvent:GamepadTouch| be a {{GamepadTouch}}.
+                  </li>
+                  <li>Set |touchEvent|.{{GamepadTouch/surfaceId}} to be
+                    |surfaceId|.
+                  </li>
+                  <li>If the touch data is part of an existing 
+                    [=active touch point=]
+                    tracked by the user agent:
+                    <ol>
+                    <li>Set |touchEvent|.{{GamepadTouch/touchId}} to the touchId of the [=active touch point=].
```
nondebug: Please use [HTML Tidy](https://www.html-tidy.org/) to update the formatting in this file.

This command will fix formatting errors in extensions.html, overwriting the input file:

`tidy -config tidyconfig.txt -m extensions.html`

extensions.html
```diff
@@ -50,7 +50,7 @@
                 company: "Google", companyURL: "http://www.google.com/",
                 w3cid: 87824 },
           ],
-
+          xref: true,
```
nondebug: Please remove this `xref` in favor of the `xref` below with a list of specification short names.

nondebug (commented):
The extensions.html content looks good to me, I think this can be merged once the formatting issues are fixed.

extensions.html
bradleyneedham: Done

nondebug (approved):
lgtm

marcoscaceres:
@bradleyneedham, would you mind rebasing so it's easier to review?

bradleyneedham:
Hi @marcoscaceres, took me a little but I think I got the rebase correct. This will hopefully be easier to see the changes.

extensions.html
```diff
@@ -1,920 +1,1152 @@
...
+        <p>
+          In addtion to the steps defined in the main <a href=
+          "https://w3c.github.io/gamepad/">Gamepad specification</a>. When the
+          user agent modifies the list of |active touch points| for a touch
+          surface by adding touch points, removing touch points, or updating
+          the values of existing touch points, <dfn data-lt=
+          "updating touchEvents">update touchEvents</dfn> by running the
+          following steps:
+        </p>
+        <ol>
```
```suggestion
        <ol class="algorithm">
```

extensions.html
```diff
@@ -1,920 +1,1152 @@
...
+          <td>
+            The current pose of the gamepad.
+          </td>
+        </tr>
+        <tr>
+          <td>
+            <dfn data-dfn-for="Gamepad">[[\touchEvents]]</dfn>
+          </td>
+          <td>
+            <code>null</code>
```
```suggestion
            `null`
```

marcoscaceres: I wonder if this should initially just be an empty list? Then we don't need to deal with list and `null` as types

extensions.html
```diff
@@ -1,920 +1,1152 @@
...
+          </td>
+        </tr>
+      </table>
+      <dl data-dfn-for="Gamepad">
+        <dt>
+          <dfn>hand</dfn>
+        </dt>
+        <dd>
+          An enumeration that indicates which hand the controller is being held
+          in or is most likely to be held in. `"left"`, `"right"`, or `""`.
```
marcoscaceres: Would "both" be better than ""?

extensions.html
bradleyneedham: Updated.

extensions.html
bradleyneedham: The enumeration is better described above in the document. I changed this to refer to the above definition. I think "" can mean "both", "either", or "NA", and perhaps it only matters to the developer if it is "left" or "right".
Perhaps @nondebug  has more insight.

extensions.html
bradleyneedham: I think because not all Gamepads support touch pads, it is better this way to distinguish between a Gamepad that will never have touchIds (null) and a Gamepad that just has no new input (empty list).

bradleyneedham:
@nondebug @sagoston @marcoscaceres Thank you all again for your review and comments. 
I rebased this, it still looks good to merge, the error that is now showing is in "index.html" not a file I changed. Let me know if there is something else I need to fix.

bradleyneedham:
@marcoscaceres I have added links to the explainer, tests, and standards positions in the first comment. Thank you again for your time and consideration.

bradleyneedham:
Hi @nondebug, the failure I am getting here is the same as the failure that https://github.com/w3c/gamepad/pull/172 is getting.
It has to do with the reference to `[="xr-standard" gamepad mapping=]` on line [1023-1024 of `index.html`](https://github.com/w3c/gamepad/blob/bb37e813b7c1f627b698a99ba49e303198f97025/index.html#L1023C73-L1024C73).
The reference appears to be correct but the test does not appear to like it.

nondebug:
Relevant error message:

> Error: ROR] Couldn’t find “&quot;xr-standard&quot; gamepad mapping“ in this document or other cited documents: [dom], [hr-time], [html], [infra], [permissions-policy], [webidl], and [webxr-gamepads-module].

I'm not sure why that reference fails, it appears to be correct according to the xref tool:

https://respec.org/xref/?term=%22xr-standard%22+gamepad+mapping

@marcoscaceres can you take a look?

extensions.html
```diff
@@ -95,8 +95,7 @@
           // document unless you know what you're doing. If in doubt ask your friendly neighbourhood
           // Team Contact.
           wgPatentURI:  "https://www.w3.org/2004/01/pp-impl/83482/status",
-
-          xref: ["DOM", "HTML", "INFRA", "gamepad"],
+          xref: ["HTML", "DOM", "PERMISSIONS-POLICY", "HR-TIME", "INFRA", "gamepad"],
```
dontcallmedom, line 98:
```suggestion
          xref: ["HTML", "DOM", "PERMISSIONS-POLICY", "HR-TIME", "INFRA", "gamepad","webxr-gamepads-module"],
```
This should solve the ReSpec/xref issue

nondebug: I think that's not sufficient to solve the issue. The failure is in index.html in a part of the document not modified by the PR, so I suspect it was a change in respec that caused this failure.

@marcoscaceres said it looks like a respec bug. The xref sees \&quot; instead of double-quote characters so it fails to match against the webxr-gamepads-module reference.

PR #136, opened 2020-06-18
Editorial: add a Model section

nondebug:
Refactor is editorial and does not add any normative requirements.

***
<a href="https://pr-preview.s3.amazonaws.com/nondebug/gamepad/pull/136.html" title="Last updated on Aug 8, 2024, 11:52 PM UTC (a51eb0f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/136/471b047...nondebug:a51eb0f.html" title="Last updated on Aug 8, 2024, 11:52 PM UTC (a51eb0f)">Diff</a>

PR #172, opened 2022-12-08
Update getGamepads to include steps for copying gamepads

nondebug:
Closes #???

Implementation commitment:

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/pull/172.html" title="Last updated on Sep 27, 2024, 12:19 AM UTC (8b1dc9a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/172/576061a...8b1dc9a.html" title="Last updated on Sep 27, 2024, 12:19 AM UTC (8b1dc9a)">Diff</a>


marcoscaceres (approved):
This is an improvement, but I don't think this is exactly right... I think we should use #136 here..

I think what we actually want is:

* The model says there is a *conceptual gamepads* 🎮🎮🎮 representing real (or conceptual) things. 
* When .getGamepads() is called, we go to the model and "snapshot" each conceptual gamepad (i.e., make the lists, copy stuff from the conceptual ones). 
* return the result. 

Right now, this reads like we are recycling what is normally returned by `.getGamepads()`.  

I'm approving, but let's move to using the model.

PR #180, opened 2023-02-09
Specify milliseconds for duration and startDelay

### LINKED ISSUE #177: Specify that GamepadEffectParameters.startDelay is a value in milliseconds
The specification currently indicates that the `duration` parameter is in milliseconds but doesn't explicitly say so for `startDelay`. I think it would be nice to clarify.

nondebug:
Closes #177

Implementation commitment:

### DIFF
--- a/extensions.html
+++ b/extensions.html
@@ -613,13 +613,13 @@
               interval, the actuator must not vibrate. An effect with negative
               {{GamepadEffectParameters/startDelay}} is not a [=valid effect=].
               If the {{GamepadEffectParameters/startDelay}} is not specified,
-              there is no initial delay before vibration starts.
+              the default delay is 0 milliseconds.
             </p>
             <p>
               {{GamepadEffectParameters/duration}} sets the duration in
               milliseconds that the actuator will vibrate. If
-              {{GamepadEffectParameters/duration}} is not specified the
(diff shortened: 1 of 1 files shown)


extensions.html
```diff
@@ -606,18 +606,20 @@ <h3>
...
             <p>
               {{GamepadEffectParameters/startDelay}} sets the duration of the
               delay after {{GamepadHapticActuator/playEffect()}} is called
-              until vibration is started, default 0. During the delay interval,
-              the actuator must not vibrate. An effect with negative
+              until vibration is started, in milliseconds. During the delay
+              interval, the actuator must not vibrate. An effect with negative
               {{GamepadEffectParameters/startDelay}} is not a [=valid effect=].
+              If the {{GamepadEffectParameters/startDelay}} is not specified,
+              there is no initial delay before vibration starts.
```
marcoscaceres: I wonder if it's clearer to treat this (no initial delay) as 0? Or am I misreading?

marcoscaceres (approved):
Just a question...

extensions.html
nondebug: It's probably best to describe it as a default value. Defining a default value makes it clear that undefined startDelay should behave the same as startDelay == 0.

PR #182, opened 2023-02-09
Update playingEffectPromise synchronously

### LINKED ISSUE #179: GamepadHapticActuator.playEffect() algorithm seems wrong with regards to `[[playingEffectPromise]]`
GamepadHapticActuator.playEffect() algorithm seems wrong with regards to `[[playingEffectPromise]]`:
- https://w3c.github.io/gamepad/extensions.html#dom-gamepadhapticactuator-playeffect

Step 4 says to *queue a task* to set `[[playingEffectPromise]]` to undefined. Step 6 says to let `[[playingEffectPromise]]` be a new promise. I believe The queued in task 4 would thus clear the new promise instead of the one from the previously playing effect.

nondebug:
Closes #179

Implementation commitment:

### DIFF
--- a/extensions.html
+++ b/extensions.html
@@ -277,17 +277,18 @@
             [=a promise resolved with=] {{GamepadHapticsResult/"preempted"}}.
             </li>
             <li>If [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}}
-            is not `undefined`, [=queue a global task=] on the [=relevant
-            global object=] of [=this=] using the [=gamepad task source=] to
-            run the following steps:
+            is not `undefined`:
               <ol>
-                <li>[=Resolve=]
-                [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}}
-                with {{GamepadHapticsResult/"preempted"}}.
+                <li>Let |effectPromise| be
+                [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}}.
                 </li>
                 <li>Set
                 [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}} to
                 `undefined`.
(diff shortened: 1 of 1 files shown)


extensions.html
```diff
@@ -277,18 +277,19 @@ <h2>
...
+                [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}}.
                 </li>
                 <li>Set
                 [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}} to
                 `undefined`.
                 </li>
+                <li>[=Queue a global task=] on the [=relevant global object=]
+                of [=this=] using the [=gamepad task source=] to [=resolve=]
+                |effectPromise| with {{GamepadHapticsResult/"preempted"}}.
+                </li>
```
marcoscaceres, line 292: This code path ends with promise being resolved, right? 

```suggestion
                </li>
                <li>Terminate this algorithm.</li>
```

extensions.html
```diff
@@ -354,18 +355,19 @@ <h2>
...
```
marcoscaceres, line 370: As above?? 

```suggestion
                </li>
                <li>Terminate this algorithm.</li>
```

marcoscaceres (approved):

extensions.html
nondebug: This path is for preempting the currently playing effect to prepare for the new effect. So there are two promises, the old effect promise and the new effect promise. We want to resolve the first promise (if it exists) and then create the new one. If we terminate here then we won't start the new effect.

extensions.html
nondebug: If we terminate here then we won't stop vibration when there's an effect playing.

Issue #148, opened 2021-04-03
GamepadHapticActuator pulse should support frequency parameter


davehill00:
Many haptic actuators support both pulsing at different frequencies, which provides a wider range of expressiveness in haptic feedback. The pulse function could easily be extended to support a 0..1 frequency value as well (which the UA could easily ignore if the hardware doesn't support different frequencies).

marcoscaceres:
Thanks @davehill00! We will revise your proposal later this week.

davehill00:
Thanks, @marcoscaceres! We're interested in adding support for this to Oculus Browser.

nondebug:
The GamepadHapticActuator definition in the Gamepad Extensions spec is out of date and I don't think it was ever fully implemented in any browser. Firefox has IDL for it but as far as I can tell none of the platform backends ever create any actuators.

Chromium implements the haptics API described in the [pull request here](https://github.com/w3c/gamepad/pull/68). It's designed for DualShock-style haptics as opposed to the pulse effect in the original GamepadHapticActuator spec. DualShock controllers have two eccentric rotating mass (ERM) actuators with differing masses that can be run at different speeds to create rumble effects. The haptics element responsible for vibrating the whole controller is exposed as `Gamepad.vibrationActuator`. Example usage:

    let gamepads = navigator.hid.getGamepads();
    if (gamepads[0] && gamepads[0].vibrationActuator) {
      gamepads[0].vibrationActuator.playEffect(
          'dual-rumble',  // Effect type
          {duration:1000,  // Vibration duration, in milliseconds
           startDelay:0,  // Delay before starting the effect, in milliseconds
           strongMagnitude:0.5,  // Vibration intensity of the "strong" (low frequency) rumble motor
           weakMagnitude:1.0});  // Vibration intensity of the "weak" (high frequency) rumble motor
    }

Currently the spec only defines one type of effect. I think the best user experience occurs when applications can use a single effect type and browsers do their best to support that effect type across all gamepads, even if the gamepad does not have the correct hardware to support that type of effect. For instance, Chrome supports sending `dual-rumble` effects to a Switch Joy-Con even though the Joy-Con has a linear resonant actuator (LRA) instead of the dual-ERM configuration. This approach might also be appropriate for Oculus VR controllers.

[...]

Issue #138, opened 2020-07-20
Xbox One impulse trigger effects


nondebug:
The haptic vibration proposal (https://github.com/w3c/gamepad/pull/68) only defines the typical "dual-rumble" style rumble effect with strong/weak ERM actuators. Xbox One controllers additionally have [Impulse Triggers](https://www.pcgamingwiki.com/wiki/Controller:Xbox_One_Controller#Impulse_Trigger_Vibration): haptic actuators that can be used to create effects localized to a particular trigger. To support these let's add new actuator types, for instance:

        enum GamepadHapticActuatorType {	
          "dual-rumble",
          "impulse-trigger-left",  // NEW
          "impulse-trigger-right",  // NEW
        };

Xbox One impulse triggers can be controlled on Windows 10 using the [Windows.Gaming.Input API](https://docs.microsoft.com/en-us/uwp/api/windows.gaming.input.gamepadvibration?view=winrt-19041).

marcoscaceres:
Seems pretty straight forward to add. As I'm new to the actuator discussion, what's that envisioned API shape for activating an actuator look like? Is it something like `gamepad.something("impulse-trigger-left", 0.4)` ?

nondebug:
The activation method could be the same as for `dual-rumble` effects but with a different set of parameters.

```
// 1 second dual-rumble effect (100% strong, 50% weak)
navigator.getGamepads()[0].vibrationActuator.playEffect("dual-rumble", {
    strongMagnitude: 1.0,
    weakMagnitude: 0.5,
    duration: 1000
});

// 1 second impulse trigger effect, left trigger (40% magnitude)
navigator.getGamepads()[0].vibrationActuator.playEffect("impulse-trigger-left", {
    magnitude: 0.4,
    duration: 1000
});
```

It might make sense to also have a combined `impulse-trigger` effect to make it easier to send effects to both triggers at the same time.

```
// 1 second impulse trigger effect, both triggers (10% left, 20% right)
navigator.getGamepads()[0].vibrationActuator.playEffect("impulse-trigger", {
    leftMagnitude: 0.1,
    rightMagnitude: 0.2,
    duration: 1000
});
```

The `vibrationActuator` object should also support querying its capabilities so applications can determine if impulse trigger effects are supported.

```
-> navigator.getGamepads()[0].vibrationActuator.effects
<- [ "dual-rumble", "impulse-trigger", "impulse-trigger-left", "impulse-trigger-right" ]
```

[...]

PR #190, opened 2023-11-14
Add vibrationActuator to the main spec

### LINKED ISSUE #173: move vibrationActuator to main spec
It be great if we could move vibrationActuator to the main spec.

gabrielsanbrito:
Closes #173

Implementation commitment:

 * [x] WebKit (https://bugs.webkit.org/show_bug.cgi?id=267523) - implemented and shipping
 * [x] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=) - implemented and shipping. 
 * [ ] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=1824217)

***
<a href="https://pr-preview.s3.amazonaws.com/gabrielsanbrito/gamepad/pull/190.html" title="Last updated on Jan 18, 2024, 12:54 AM UTC (2512d16)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/190/775b8e1...gabrielsanbrito:2512d16.html" title="Last updated on Jan 18, 2024, 12:54 AM UTC (2512d16)">Diff</a>


gabrielsanbrito:
@nondebug, @marcoscaceres Just a few questions I had while editing:
- Should we remove the GamepadHapticActuatorType from the spec, as it has been marked as deprecated? If that is true, should we remove the construction step for the GamepadHapticActuator interface?
  - If we are keeping it, i have made minor changes to the definitions of "vibration" and "dual-rumble" actuator types to fix a CI error where `Gamepad.hapticActuators` was being mentioned in the main spec without being previously defined.

- I have deleted the "This specification defines an external now hidden algorithm [...]" sentence on the GamepadHapticActuator "Page visibility" subsection. Should we create a pull request to add the visibility changed steps to the HTML spec(https://html.spec.whatwg.org/multipage/interaction.html#page-visibility), like the Screen Orientation API has done?

nondebug:
> Should we remove the GamepadHapticActuatorType from the spec

Yes, let's remove it. It was never really part of the spec so it's not even officially deprecated.

> Should we create a pull request to add the visibility changed steps to the HTML spec

Yes, let's do that.

index.html
```diff
@@ -1,1688 +1,2240 @@
...
+        </tr>
+        <tr>
+          <td>
+            <dfn data-dfn-for="Gamepad">[[\vibrationActuator]]</dfn>
+          </td>
+          <td>
+            null
+          </td>
+          <td>
+            A {{GamepadHapticActuator}} object capable of generating a haptic effect that vibrates the entire gamepad.
```
nondebug: A {{GamepadHapticActuator}} object capable of generating a haptic effect that vibrates the entire gamepad, or null if the gamepad has no such capability.

index.html
```diff
@@ -1,1688 +1,2240 @@
...
+          <p>
+            The {{Gamepad/buttons}} getter steps are:
+          </p>
+          <ol>
+            <li>Return [=this=].{{Gamepad/[[buttons]]}}.
+            </li>
+          </ol>
+        </dd>
+        <dt>
+          <dfn>vibrationActuator</dfn> attribute
```
nondebug: We should define algorithmic steps for the vibrationActuator attribute getter like we do for other attributes.

```
<p>
  The {{Gamepad/vibrationActuator}} getter steps are:
</p>
<ol>
  <li>Return [=this=].{{Gamepad/[[vibrationActuator]]}}.
  </li>
</ol>
```

I think the paragraph below is no longer needed since we now define steps for initializing the internal slot in [=initialize the vibration actuator=].

index.html
```diff
@@ -1,1688 +1,2240 @@
...
+      </h2>
+      <p>
+        A {{GamepadHapticActuator}} corresponds to a configuration of motors or
+        other actuators that can apply a force for the purposes of haptic
+        feedback.
+      </p>
+      <pre class="idl">
+        [Exposed=Window]
+        interface GamepadHapticActuator {
+          readonly attribute GamepadHapticActuatorType type;
```
nondebug: Let's remove `GamepadHapticActuatorType type`, it is not needed now that we have `canPlayEffectType`.

index.html
```diff
@@ -1,1688 +1,2240 @@
...
+          boolean canPlayEffectType(GamepadHapticEffectType type);
```
nondebug: In an earlier review [Marcos suggested](https://github.com/w3c/gamepad/pull/163#pullrequestreview-1000657433) some modifications to `canPlayEffectType`. I think there are no implementations yet so we are unconstrained.

suggestion 1: rename to "canPlay" or "supports"
suggestion 2: change the parameter to a DOMString so it's not an error to pass a new effect type to an old browser that doesn't recognize it
suggestion 3: replace "canPlayEffectType" with an "effects" attribute that returns a FrozenArray of supported effect types

Suggestion 3 seems best to me.

index.html
```diff
@@ -1,1688 +1,2240 @@
...
+                  </ol>
+                </li>
+              </ol>
+            </li>
+            <li>Return {{GamepadHapticActuator/[[playingEffectPromise]]}}.
+            </li>
+          </ol>
+        </dd>
+        <dt>
+          <dfn>pulse()</dfn> method
```
nondebug: We should define steps for `pulse`.

index.html
```diff
@@ -1,1688 +1,2240 @@
...
+        [Exposed=Window, SecureContext]
+        interface Gamepad {
+          readonly attribute DOMString id;
+          readonly attribute long index;
+          readonly attribute boolean connected;
+          readonly attribute DOMHighResTimeStamp timestamp;
+          readonly attribute GamepadMappingType mapping;
+          readonly attribute FrozenArray&lt;double&gt; axes;
+          readonly attribute FrozenArray&lt;GamepadButton&gt; buttons;
+          [SameObject] readonly attribute GamepadHapticActuator? vibrationActuator;
```
nondebug: To support `pulse()` we probably need to bring back the `hapticActuators` array:

```
readonly attribute FrozenArray&lt;GamepadHapticActuator> hapticActuators;
```

With `pulse()` it's expected that a gamepad could have multiple haptic actuators and each actuator can be controlled independently. With `playEffect()` the `vibrationActuator` member is expected to control multiple actuators. This means there should be overlap between the actuators in `hapticActuators` and `vibrationActuator`. The spec should explain how to handle pre-empting `pulse()` with `playEffect()` and vice versa.

nondebug (changes requested):

index.html
gabrielsanbrito: When trying to write this, I felt like it is a little bit confusing to have both  a `FrozenArray<GamepadHapticActuator> hapticActuators` and a `GamepadHapticActuator vibrationActuator` on the the Gamepad interface. For example, to keep consistency, if we decide to also expose `hapticActuators`, should the two dual-rumble motors also be individually exposed as 2 extra elements of the `hapticActuators` array? If that is the case, why should we do it when we can achieve the same level of control by calling playEffect? We can activate only one of the motors with:
```js
gamepad.vibrationActuator.playEffect("dual-rumble", {
        duration: 1000,
        strongMagnitude: 0.5,
 });
```

As mentioned in #148, the Quest Controller Pro has 3 haptic actuators (one in the grip, one in the thumb rest, and another in the trigger). I see that the `pulse()` proposal was made some time ago already. Maybe, this type of controller could take advantage of the trigger rumble extension proposed earlier (#138). Another issue that I see with the `hapticActuators` approach is that there is no standardized way (yet) for us to identify the actuators (which one is the left trigger? which one is the right ERM?) and actuate the desired one. 

For now, I would suggest to postpone adding the `hapticActuators` array and the `pulse()` method to the Gamepad spec and, instead, keep `pulse()` in the Gamepad Extensions document inside a partial GamepadHapticActuator Interface. Later on we can discuss further how to support both `hapticActuators` and `pulse()`. What do you think?

index.html
gabrielsanbrito: Added suggestion 3 to the spec.

index.html
nondebug: > to keep consistency, if we decide to also expose hapticActuators, should the two dual-rumble motors also be individually exposed as 2 extra elements of the hapticActuators array?
 
I think we should expose dual-rumble actuators as 2 elements in `hapticActuators` as well as a non-null `vibrationActuator`.

> there is no standardized way (yet) for us to identify the actuators (which one is the left trigger? which one is the right ERM?) and actuate the desired one

The standard can be opinionated on actuator indices like it is for button and axis indices. If we follow the Standard Gamepad example then we should define standard actuators based on layout. That said, we generally can't detect layout programmatically for "dual-rumble" gamepads. APIs usually label them as strong/weak without specifying where they are located. I don't think the standard should care where they are located either since the intent is to vibrate the whole gamepad, not one particular side of it. Still, we need some way for apps to distinguish them.

I would be fine with using the same indices that Android uses (index 0 for strong vibration channel, index 1 for weak vibration channel). I think it makes sense to give these special placement since they are the most common type of gamepad haptic effect.

For gamepads with actuators that aren't intended to vibrate the entire gamepad (like Xbox trigger rumble or Quest Controller Pro) we should probably start thinking about adding GamepadHapticActuator attributes to identify placement instead of relying on standard indices. Perhaps we can lean on OpenXR's concept of semantic paths to describe actuator placement in a generalizable way.

https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#semantic-path

> why should we do it when we can achieve the same level of control by calling playEffect?

`pulse()` should not preempt effects playing on other actuators. With `playEffect()` apps need to issue a new effect each time either actuator's vibration should change. This is fine for dual-rumble but I think it might be cumbersome for trigger-rumble and other actuator configurations.

> For now, I would suggest to postpone adding the hapticActuators array and the pulse() method to the Gamepad spec and, instead, keep pulse() in the Gamepad Extensions document inside a partial GamepadHapticActuator Interface. Later on we can discuss further how to support both hapticActuators and pulse(). What do you think?

That's fine with me, we can focus on merging `vibrationActuator` and `playEffect()` for now to unblock trigger-rumble. There aren't any implementations yet that implement both `vibrationActuator` and `hapticActuators` so it's probably premature to define their interaction in the main spec. If we keep `hapticActuators` and `pulse()` in the extensions doc then we can iterate on them there.

gabrielsanbrito: Great! Thanks for your always super detailed answers! I have created a partial GamepadHapticActuator interface in the extensions document.

index.html
gabrielsanbrito: Pulse has been moved back to the extensions document inside the partial GamepadHapticActuator interface.

marcoscaceres (changes requested):
Um, any idea why it's showing both files as being entirely changed? I think you might have changed the line endings or something? (i.e., this makes it not possible to review the changes)

Can you try sending it again or reach out to us if you need a hand? Are you doing this on windows machine maybe?

marcoscaceres:
Yeah, @gabrielsanbrito, it seems you've accidentally changed the whitespace of the file:
https://github.com/w3c/gamepad/pull/190/files?diff=unified&w=1

Can you see about changing it back? 

Maybe try:
```
git rebase --whitespace=fix gh-pages
```

extensions.html
```diff
@@ -1145,6 +617,38 @@ <h3>
...
+        partial interface GamepadHapticActuator {
+          Promise&lt;boolean&gt; pulse(double value, double duration);
+        };
+      </pre>
+      <dt>
+        <dfn>pulse()</dfn> method
+      </dt>
+      <dd>
+        <p>
+          <code>pulse()</code> applies a <var>value</var> to the actuator for
```
```suggestion
          {{GamepadHapticActuator/pulse()}} applies a |value:double| to the actuator for
```

extensions.html
```diff
@@ -1145,6 +617,38 @@ <h3>
...
+          <var>duration</var> milliseconds. The <var>value</var> passed to
+          <code>pulse()</code> is clamped to limits defined by the actuator
+          type. The returned Promise will resolve <code>true</code> once the
+          pulse has completed.
+        </p>
+        <p>
+          Repeated calls to <code>pulse()</code> override the previous
```
marcoscaceres, line 647: What does override mean here? What happens to the p1?

```
p1  = ...pulse();
p2  = ...pulse()
```

Does p1 get rejected?

index.html
```diff
@@ -153,6 +153,7 @@ <h2>
           readonly attribute GamepadMappingType mapping;
           readonly attribute FrozenArray&lt;double&gt; axes;
           readonly attribute FrozenArray&lt;GamepadButton&gt; buttons;
+          [SameObject] readonly attribute GamepadHapticActuator? vibrationActuator;
```
marcoscaceres: Why is it nullable? that seems a bit odd?

extensions.html
gabrielsanbrito: I didn't write this part of the spec. This `pulse()` definition is already in the [Gamepad Extensions spec](https://w3c.github.io/gamepad/extensions.html#dom-gamepadhapticactuator-pulse). Although, my understanding is that it is not clear whether `p1` would be rejected or resolved with `false`. I think that by "override" the writer meant that any currently playing pulse effect (`p1`) should be immediately preempted, and `p2` should start playing right away  I think that it should look something like:
- If `p1` is preempted, resolve with `true`, because it was playing anyways;
- If `p1` didn't start playing at all, resolve with `false`
- If `pulse` is not supported for any reason, reject the promise with an error.

However, since this exact wording is already in the extensions document, I would suggest leaving it as it is right now to focus on moving `vibrationActuator` to the main spec. Later on we can discuss further on how to better specify `pulse` and the `hapticActuators` array. What do you think?

index.html
gabrielsanbrito: I don't know. It makes sense to me at least. I have used a few gamepad devices in the past that didn't have any sort of vibration actuators. It makes sense to me to have it being `null` instead of having an empty `GamepadHapticActuator.effects` array. What do you think?

extensions.html
gabrielsanbrito: Pushed the fix. Thanks!

extensions.html
nondebug: I think the only implementation of `pulse` is in Meta Quest Browser. We should find out how it behaves and write the spec to match that implementation as closely as possible.

I looked at the original PR (https://github.com/w3c/gamepad/pull/32) and didn't see any discussion of whether the promise returned by pulse should ever do anything but resolve to true. Maybe @toji remembers?

Personally, I would want the promise returned by `pulse()` to resolve to true only if the effect played for the entire duration without being pre-empted. I expect apps will want to chain multiple calls to `pulse()` to make more complex effects. If one of the pulses is pre-empted, apps should be able to detect that and avoid playing the subsequent chained pulses.

So, I would say `p1` should be resolved with false after the second call to `pulse()`. We shouldn't reject in this case because pre-empting an effect is not an error.

index.html
nondebug: The Chromium implementation made it nullable for feature detection. Null `vibrationActuator` means no vibration capability.

It's odd that we now have two levels of feature detection and thus two ways to represent a gamepad with no vibration capability. To make feature detection simpler, the spec should pick one representation.

We should avoid breaking existing apps that use the null-check for feature detection. Making `vibrationActuator` non-nullable would cause these apps to play effects on gamepads that don't support any effects. That might be okay depending on how unsupported effects are handled.

According to the spec, `playEffect` called with an unsupported effect type returns a promise rejected with NotSupportedError. The actual behavior in the Chromium implementation is to resolve the promise with "not-supported". My guess is applications aren't checking for either of these. In the no-vibration-capability case, unexpectedly rejecting with NotSupportedError is probably more disruptive to existing applications than unexpectedly resolving with "not-supported". I think it would be reasonable to make `vibrationActuator` non-nullable if we also change `playEffect` to resolve to "not-supported" in this case.

If we decide to keep `vibrationActuator` nullable then the spec should require it to be null whenever `effects` would be empty.

marcoscaceres: supporting null seems a bit iffy to be honest (and kinda a pain to implement, because you need to know if you can play effects before you bind the object). 

My preference would be to make it non-nullable, as it's still an expectation that canPlayEffectType() can still return not-supported for everything. 

Also, the actual feature detection check is if it's `undefined` right now for browsers that don't implement this. 

Yes, there is risk of breakage for browsers, but developers still need to check (fake code for example):

```JS
// is supported at all
if (gamepad.vibrationActuator !== undefined) {
   // it's implemented, but can't play anything
    if (gamepad.vibrationActuator !== null) {
      // can it play the thing I need
      if (gamepad.vibrationActuator.canPlayEffectType(type)) {
      }
}
```

Which seems excessive. It should just be regular (which will work regardless):

```JS
if (gamepad.vibrationActuator?.canPlayEffectType(type)) {
   // play it...
}
```

marcoscaceres: I guess put differently, Chrome could continue to support null, but maybe we shouldn't codify that in the actual spec?

extensions.html
marcoscaceres: @cabanier ☝️...

marcoscaceres: Ok, but if pulse() is Meta Quest only, that also puts it the "feature at risk". We should note that in the spec too, unless we expect to get a second implementation. 

Is there a plan for a second implementation?

marcoscaceres: Actually, this is basically an incubation, so we don't need to block on this. 

But we should figure out if anyone will actually support pulse() or not and if not, we should consider dropping it.

index.html
```diff
@@ -851,6 +885,20 @@ <h3>
           <li>Return |buttons|.
           </li>
         </ol>
+        <p>
+          To <dfn data-lt="initializing the vibration actuator">initialize the
+          vibration actuator</dfn> for a gamepad, run the following steps:
+        </p>
+        <ol>
+          <li>If |gamepad:Gamepad| has a vibration actuator that can
```
marcoscaceres: This seems overly specific... I know we might not change this in the future, but making it depend on "dual-rumble" exclusively doesn't seem great. This should probably just be that it it supports haptic effects.

extensions.html
toji: Broadly agree with @marcoscaceres that we should evaluate whether or not `pulse()` is actually desired going forward. I'll also throw up the @cabanier bat signal here to get Meta's take on it.

Re: the intent behind the `pulse()` promise resolution...
84_Years

It _has_ been a while, so I don't recall the details of the conversation around this API. But I do recall having conversations with the TAG in regards to similarly shaped APIs at around the same time. Specifically the [WebXR `isSessionSupported()` API](https://immersive-web.github.io/webxr/#xrsystem-interface), which also resolves to a boolean value. 

Originally we were going to have the promise resolve with void if the session type was supported and reject otherwise but the guidance we received was that rejection should only happen in the case of something truly exceptional, where the intent of the method called could not be fulfilled. This is especially true when you consider that many promises are used with async/await patterns, and the only way to observe the rejection is a try/catch. That's fairly disruptive. In the case of `isSessionSupported()` the question being asked (can I start this kind of session) could always be answered yes or no, so resolving with a boolean made sense and the method never rejects.

With that framing in mind, the fact that this resolves to a boolean rather than void tells me that the boolean was intended to communicate information. It feels to my like the intent was "resolve when the requested pulse is finished, with true if it was allowed to complete and false if it was pre-empted." I wouldn't expect it to reject unless perhaps the pulse simply couldn't be rendered by the device haptics at all? Even then it seems possibly more ergonomic to never reject.

Whether or not that's what the Meta implementation does is an entirely different story.

marcoscaceres: yeah, I agree that not rejecting makes sense. 

I'm still hopeful pulse can go away though... unless it's standardized across different controllers, it's still quite ambiguous about what it actually means or how it would be represented/felt (duration, strength, amplitude, etc.).

index.html
gabrielsanbrito: I will make the `vibrationActuator` non-nullable in the next commit.

gabrielsanbrito: I have made it non-nullable

index.html
gabrielsanbrito: I have removed the mention to dual rumble and also added construction steps for GamepadHapticActuator.

extensions.html
gabrielsanbrito: What do you think that we should do now? Should we remove it or keep it in the extensions document? I am fine with keeping it in the extensions document until we can reach consensus.

marcoscaceres: yeah, let's keep it in the extensions document.

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+        </dd>
+        <dt>
+          <dfn>playEffect()</dfn> method
+        </dt>
+        <dd>
+          <p>
+            The {{GamepadHapticActuator/playEffect()}} method steps are:
+          </p>
+          <ol>
+            <li>If |params:GamepadEffectParameters| does not describe a [=valid
```
marcoscaceres, line 1155: I think what we actually want here is a switch statement for performing validation of each `type` and a give `params`. Right now, it's you need to go hunting for what do do, which is not great.

I was expecting:
```HTML 
<p>
The {{GamepadHapticActuator/playEffect()}} method steps, called with {{GamepadHapticEffectType}} 
|type:GamepadHapticEffectType| and {{GamepadEffectParameters}} |params:GamepadEffectParameters | :
</p>
<ol class="algorithm">
  <li>Given the value of type |type:GamepadHapticEffectType|, switch on:
  <dl class="switch">  
    <dt>
      {{GamepadHapticEffectType/"dual-rumble"}}
    </dt>
    <dd>
      <ol>
        <li>let |result be [=validate "dual-rumble" parameters=] passing |params|.
        <li>If result is an an error, return [=a promise rejected=] with error. 
      <ol>
    </dd>
  </dl>
</li>
<!--- other steps --> 
</ol>
```

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+            effect=] of type |type:GamepadHapticEffectType|, return [=a promise
+            rejected with=] reason {{TypeError}}.
+            </li>
+            <li>Let |document| be the [=current settings object=]'s [=relevant
```
```suggestion
            <li>Let |document:Document?| be the [=current settings object=]'s [=relevant
```

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+                [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}} to
+                `undefined`.
+                </li>
+                <li>[=Queue a global task=] on the [=relevant global object=]
+                of [=this=] using the [=gamepad task source=] to [=resolve=]
+                |effectPromise| with {{GamepadHapticsResult/"preempted"}}.
+                </li>
+              </ol>
+            </li>
+            <li>If this actuator cannot [=play effects with type=] |type|,
```
```suggestion
            <li>If |this|'s gamepad's actuator cannot [=play effects with type=] |type|,
```

marcoscaceres: I wonder if this check should happen earlier?

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+            return [=a promise rejected with=] reason {{NotSupportedError}}.
+            </li>
+            <li>Let {{GamepadHapticActuator/[[playingEffectPromise]]}} be [=a
+            new promise=].
+            </li>
+            <li>Do the following steps in parallel:
```
```suggestion
            <li>Do the following steps [=in parallel=]:
```

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+              <ol>
+                <li>[=Issue a haptic effect=] to the actuator with |type| and
+                |params|.
+                </li>
+                <li>When the effect completes, if it was not preempted, [=queue
+                a global task=] on the [=relevant global object=] of [=this=]
+                using the [=gamepad task source=] to run the following steps:
+                  <ol>
+                    <li>[=Resolve=]
```
marcoscaceres, line 1209: Should this assert or check that [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}} is not undefined? It could have gotten destroyed or it might not even be the same promise anymore, right?

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+        </dd>
+        <dt>
+          <dfn>reset()</dfn>
+        </dt>
+        <dd>
+          <p>
+            The {{GamepadHapticActuator/reset()}} method steps are:
+          </p>
+          <ol>
+            <li>Let |document| be the [=current settings object=]'s [=relevant
```
```suggestion
            <li>Let |document:Document?| be the [=current settings object=]'s [=relevant
```

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+                [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}} to
+                `undefined`.
+                </li>
+                <li>[=Queue a global task=] on the [=relevant global object=]
+                of [=this=] using the [=gamepad task source=] to [=resolve=]
+                |effectPromise| with {{GamepadHapticsResult/"preempted"}}.
+                </li>
+              </ol>
+            </li>
+            <li>[=Stop haptic effects=] on this actuator.
```
```suggestion
            <li>[=Stop haptic effects=] on [=this=]'s gamepad's actuator.
```

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+            </li>
+            <li>Return [=a promise resolved with=]
```
marcoscaceres: This doesn't feel right to me... it's returning the resolved promise without getting affirmation from the OS that the call to stop was sent and succeeded. I was expecting this to work the same as playEffect().

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+        |type:GamepadHapticEffectType| if |type:GamepadHapticEffectType| can be
+        found in the {{GamepadHapticActuator/[[effects]]}} [=list=].
+      </p>
+      <p>
+        An effect with {{GamepadHapticEffectType}}
+        |type:GamepadHapticEffectType| and {{GamepadEffectParameters}}
+        |params:GamepadEffectParameters| describes a <dfn>valid effect</dfn> if
+        |params| contains all of the required parameters for |type| effects and
+        all of the parameters have valid values. Additional parameters MUST be
+        ignored.
```
marcoscaceres: I don't think we need to say this: 
```suggestion
        all of the parameters have valid values. 
```

Because what is checked is defined by the algorithm. Additional parameters will be dropped by IDL dictionary, and the ones that are optional (and ignored) are also defined by the validation steps.

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+      </p>
+      <p>
+        A [=user agent=] that implements this API MUST provide a method to
```
marcoscaceres: This isn't right... The should just define should define what  `<dfn>issue a haptic effect</dfn>` with the parameters means, not that it's a MUST. The MUST part is already part of playEffect(), this just needs to say what to do "i.e., make the actuator render the |type| and try to make it use the |params|" (needs better wording, but you get the idea).

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+        other actuators that can apply a force for the purposes of haptic
+        feedback.
+      </p>
+      <pre class="idl">
+        [Exposed=Window]
+        interface GamepadHapticActuator {
+          readonly attribute FrozenArray&lt;GamepadHapticEffectType&gt; effects;
+          Promise&lt;GamepadHapticsResult&gt; playEffect(
+              GamepadHapticEffectType type,
+              optional GamepadEffectParameters params = {});
```
```suggestion
          Promise&lt;GamepadHapticsResult&gt; playEffect(
              GamepadHapticEffectType type,
              optional GamepadEffectParameters params = {}
          );
```

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+      <table class="simple">
+        <tr>
+          <th>
+            Internal slot
+          </th>
+          <th>
+            Initial value
+          </th>
+          <th>
+            Description (non-normative)
```
marcoscaceres: Looks pretty normative to me... 

```suggestion
            Description
```

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+          </th>
+        </tr>
+        <tr>
+          <td>
+            <dfn data-dfn-for="GamepadHapticActuator">[[\effects]]</dfn>
+          </td>
+          <td>
+            An empty [=list=].
```
```suggestion
            An empty [=list=] of {{Gamepad/GamepadHapticEffectType}}.
```

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+            representing all the effects supported by the actuator
+          </td>
+        </tr>
+        <tr>
+          <td>
+            <dfn data-dfn-for=
+            "GamepadHapticActuator">[[\playingEffectPromise]]</dfn>
+          </td>
+          <td>
+            `undefined`
```
marcoscaceres: I wonder if we should use `null` instead of `undefined` for this. Seems like a better fit.

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+        <tr>
+          <td>
+            <dfn data-dfn-for="GamepadHapticActuator">[[\effects]]</dfn>
+          </td>
+          <td>
+            An empty [=list=].
+          </td>
+          <td>
+            A [=list=] of {{Gamepad/GamepadHapticEffectType}} values
+            representing all the effects supported by the actuator
```
```suggestion
            Represents the effects supported by the actuator.
```

index.html
marcoscaceres: I.e., the type could be
```suggestion
            `null`
```

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+          <p>
+            The {{GamepadHapticActuator/reset()}} method steps are:
+          </p>
+          <ol>
+            <li>Let |document| be the [=current settings object=]'s [=relevant
+            global object=]'s [=associated `Document`=].
+            </li>
+            <li>If |document| is `null` or |document| is not [=Document/fully
+            active=] or |document|'s [=Document/visibility state=] is `"hidden"`, return
+            [=a promise resolved with=] {{GamepadHapticsResult/"preempted"}}.
```
marcoscaceres: This seems "exceptional". I think this should reject as it doesn't actually have the desired effect. 

Same where the other check happens.

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+        A [=user agent=] that implements this API MUST provide a method to
+        <dfn>issue a haptic effect</dfn> to an actuator with
+        |type:GamepadHapticEffectType| and |params:GamepadEffectParameters|.
+        The [=user agent=] MAY modify the effect to increase compatibility. For
+        example, an effect intended for a rumble motor may be transformed into
+        a waveform-based effect for a device that supports waveform haptics but
+        lacks rumble motors.
+      </p>
+      <p>
+        A [=user agent=] that implements this API MUST provide a method to
```
marcoscaceres: As above... the MUST is not needed here. This should just describe the intent.

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+        <dfn>stop haptic effects</dfn> on an actuator. When called, if there is
```
```suggestion
        <dfn>stop haptic effects</dfn> on an |actuator|. When called, if there is
```

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+    </section>
+    <section>
+      <h3>
+        <dfn>GamepadHapticsResult</dfn> Enum
+      </h3>
+      <p>
+        An enum representing the result of a call to
+        {{GamepadHapticActuator}}.{{GamepadHapticActuator/playEffect()}} or
+        {{GamepadHapticActuator}}.{{GamepadHapticActuator/reset()}}.
+      </p>
```
marcoscaceres: I don't think we should say this here... the definitions automatically link back to the methods, and anywhere else these enum values are used.  

```suggestion
```

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+      <dl data-dfn-for="GamepadHapticsResult">
+        <dt>
+          <dfn>complete</dfn>
+        </dt>
+        <dd>
+          <p>
+            The value the {{Promise}} returned by
+            {{GamepadHapticActuator}}.{{GamepadHapticActuator/playEffect()}}
+            or {{GamepadHapticActuator}}.{{GamepadHapticActuator/reset()}}
+            will resolve to when the effect completes normally.
```
```suggestion
            The haptic effect completed playing.
```

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+          <dfn>preempted</dfn>
+        </dt>
+        <dd>
+          <p>
+            The value the {{Promise}} returned by
+            {{GamepadHapticActuator}}.{{GamepadHapticActuator/playEffect()}}
+            or {{GamepadHapticActuator}}.{{GamepadHapticActuator/reset()}}
+            will resolve to when the operation does not complete because the
+            {{Document}} is `"hidden"` or another effect was started
+            on the same actuator.
```
```suggestion
            The effect was stopped and replaced (i.e., "preempted") by another effect.
```

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+        the actuator.
+      </p>
+      <p>
+        The [=user agent=] MAY modify the behavior of
+        {{GamepadHapticActuator}}.{{GamepadHapticActuator/playEffect}} to
+        extend support for new {{GamepadHapticEffectType}} values and
+        {{GamepadEffectParameter}} attributes. Implementations MUST NOT
+        modify the definition of a [=valid effect=] for effect types defined
+        in this specification.
+      </p>
```
marcoscaceres: New  GamepadHapticEffectType should be standardized instead. Definitely don't want to encourage support for arbitrary new types! The second clause doesn't seem to be very useful to be honest. 

```suggestion
```

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+          </p>
+          <p>
+            A {{GamepadHapticEffectType/"dual-rumble"}} effect is a
+            fixed-length, constant-intensity vibration effect intended for an
+            actuator of this type. {{GamepadHapticEffectType/"dual-rumble"}}
+            effects are defined by {{GamepadEffectParameters/startDelay}},
+            {{GamepadEffectParameters/duration}},
+            {{GamepadEffectParameters/strongMagnitude}}, and
+            {{GamepadEffectParameters/weakMagnitude}}, none of which are
+            required.
```
```suggestion
            required because they default to 0.
```

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+          </p>
+          <p>
+            {{GamepadEffectParameters/startDelay}} and
+            {{GamepadEffectParameters/duration}} set the duration of the
+            initial delay before after initiating the effect before the
+            effect will be started
+          </p>
```
marcoscaceres: This is already part of the startDelay's definition. 

```suggestion
```

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+          <p>
+            {{GamepadEffectParameters/startDelay}} sets the duration of the
+            delay after {{GamepadHapticActuator/playEffect()}} is called
+            until vibration is started, in milliseconds. During the delay
+            interval, the actuator must not vibrate. An effect with negative
+            {{GamepadEffectParameters/startDelay}} is not a [=valid effect=].
+            If the {{GamepadEffectParameters/startDelay}} is not specified,
+            the default delay is 0 milliseconds.
```
marcoscaceres: Redundant. Already stated in the WebIDL  

```suggestion
```

marcoscaceres: Why is startDelay a `double` when it doesn't allow negative and it represents milliseconds anyway? Should this just just a `unsigned long long` instead?

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+            required.
+          </p>
+          <p>
+            {{GamepadEffectParameters/startDelay}} and
+            {{GamepadEffectParameters/duration}} set the duration of the
+            initial delay before after initiating the effect before the
+            effect will be started
+          </p>
+          <p>
+            {{GamepadEffectParameters/startDelay}} sets the duration of the
```
marcoscaceres: This definition should be moved up to where `<dfn>` startDelay is defined. This isn't specific to "dual-rumble" AFAICT.

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+            {{GamepadEffectParameters/startDelay}} is not a [=valid effect=].
+            If the {{GamepadEffectParameters/startDelay}} is not specified,
+            the default delay is 0 milliseconds.
+          </p>
+          <p>
+            {{GamepadEffectParameters/duration}} sets the duration in
+            milliseconds that the actuator will vibrate. If
+            {{GamepadEffectParameters/duration}} is not specified, the
+            default duration is 0 milliseconds.
+          </p>
```
marcoscaceres: Same. And also as above.... a `double` seems like the wrong type here. Do we ever want a negative duration? Does that even make sense? 

```suggestion
```

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+          <p>
+            {{GamepadEffectParameters/strongMagnitude}} and
+            {{GamepadEffectParameters/weakMagnitude}} set the vibration
+            intensity levels for the low-frequency and high-frequency rumble
+            motors, normalized to the range `[0,1]`, default 0. An effect
+            with magnitude values outside of this range is not a [=valid
+            effect=].
```
marcoscaceres: As above... this should be generalized... it's not specific to "dual-rumble" 

```suggestion
            {{GamepadEffectParameters/strongMagnitude}} and
            {{GamepadEffectParameters/weakMagnitude}} set the 
            intensity levels for the low-frequency and high-frequency vibrations, 
            normalized to the range `[0,1]`, defaulting to 0. An effect
            with magnitude values outside of this range is not a [=valid
            effect=].
```

The " An effect with magnitude values outside of this range is not a [=valid effect=]" should be checked as part of a valuation algorithm.

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+      <p>
+        To mitigate unwanted long-running effects, the [=user agent=] MAY
+        limit the total effect duration for a [=valid effect=] to some
+        maximum duration. It is RECOMMENDED that the [=user agent=] use a
+        maximum of 5 seconds.
+      </p>
+      <pre class="idl">
+        dictionary GamepadEffectParameters {
+            double duration = 0.0;
+            double startDelay = 0.0;
```
marcoscaceres: As above... do we really want double here?

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+            </li>
+            <li>If this actuator cannot [=play effects with type=] |type|,
+            return [=a promise rejected with=] reason {{NotSupportedError}}.
+            </li>
+            <li>Let {{GamepadHapticActuator/[[playingEffectPromise]]}} be [=a
+            new promise=].
+            </li>
+            <li>Do the following steps in parallel:
+              <ol>
+                <li>[=Issue a haptic effect=] to the actuator with |type| and
```
marcoscaceres: I wonder if we should take a snapshot of a DOMHighResTimestamp before going on parallel, and then sending that along over as well. That would allow for more precise playback timing relative to when this method was called.

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+            configuration with an eccentric rotating mass (ERM) vibration
+            motor in each handle of a standard gamepad. In this
+            configuration, either motor is capable of vibrating the whole
+            gamepad. The vibration effects created by each motor are unequal
+            so that the effects of each can be combined to create more
+            complex haptic effects.
+          </p>
+          <p>
+            A {{GamepadHapticEffectType/"dual-rumble"}} effect is a
+            fixed-length, constant-intensity vibration effect intended for an
```
marcoscaceres: What does "fixed-length" mean here, and how does that play with "duration"?

extensions.html
cabanier: > Ok, but if pulse() is Meta Quest only, that also puts it the "feature at risk". We should note that in the spec too, unless we expect to get a second implementation.
> Is there a plan for a second implementation?

I'm fairly sure that this was also implemented in the Pico browser. @bkardell , was this added to Wolvic?

cabanier: > With that framing in mind, the fact that this resolves to a boolean rather than void tells me that the boolean was intended to communicate information. It feels to my like the intent was "resolve when the requested pulse is finished, with true if it was allowed to complete and false if it was pre-empted." I wouldn't expect it to reject unless perhaps the pulse simply couldn't be rendered by the device haptics at all? Even then it seems possibly more ergonomic to never reject.
> 
> Whether or not that's what the Meta implementation does is an entirely different story.

I checked the Quest implementation and it's running `pulse` through the same codepath as `playEffect` after clamping.
Looking through the code, I can't see where this would return false except for cases where the controller doesn't have haptics (ie hands)

index.html
gabrielsanbrito: I made the change, but could you please explain to me why `null` is a better fit?

index.html
gabrielsanbrito: I haven't written this, but I think that fixed-length in this context means that, once an effect with a duration `d` has been issued, you cannot modify `d`. Your only options are to either preempt it or wait for it to finish.

index.html
nondebug: I wrote it this way because we generally don't get any signal from the OS or device that the effect was actually stopped.

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+            </li>
+            <li>Let {{GamepadHapticActuator/[[playingEffectPromise]]}} be [=a
+            new promise=].
+            </li>
+            <li>Do the following steps in parallel:
+              <ol>
+                <li>[=Issue a haptic effect=] to the actuator with |type| and
+                |params|.
+                </li>
+                <li>When the effect completes, if it was not preempted, [=queue
```
marcoscaceres: Let's change "if it was not preempted" to an explicit check for `GamepadHapticActuator/[[playingEffectPromise]]`

index.html
marcoscaceres: Maybe `unsigned long long`

index.html
gabrielsanbrito: Thinking a little bit more about it, I think that only checking `GamepadHapticActuator/[[playingEffectPromise]]` might not be the best way to proceed. For example:
1. Effect `e1` starts playing. So `GamepadHapticActuator/[[playingEffectPromise]]` is set to `p1`
2. While `e1` is already executing (gamepad is already vibrating), we issue `e2`
3. `GamepadHapticActuator/[[playingEffectPromise]]` is set to `p2` and `p1` is resolved with `"preempted"`.
4. `e1` finishes playing normally and `GamepadHapticActuator/[[playingEffectPromise]]` is already set to `p2`. `p2` is resolved with `"completed"` and `GamepadHapticActuator/[[playingEffectPromise]]` is set to `null`. 

I think that we could maybe add an explicit "stop haptics effect", like we did with `reset()`:
```
<li>If [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}}
            is not `null`:
              <ol>
                <li>Let |effectPromise| be
                [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}}.
                </li>
                <li>Set
                [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}} to
                `null`.
                </li>
                <li>[=Stop haptic effects=] on [=this=]'s gamepad's actuator.
               </li>
                <li>[=Queue a global task=] on the [=relevant global object=]
                of [=this=] using the [=gamepad task source=] to [=resolve=]
                |effectPromise| with {{GamepadHapticsResult/"preempted"}}.
                </li>
              </ol>
            </li>
```
I think that is why the "if it was not preempted" part is also relevant. What do you think? Also @nondebug.

index.html
gabrielsanbrito: I have made the changes to the `reset()` method so that we wait for some platform response. The vocabulary might be a little bit different in relation to `playEffect`, but I can adjust either for consistency.

index.html
gabrielsanbrito: I have these verification steps closer to <dfn>valid effect</dfn>. WDYT?

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+            effect=] of type |type:GamepadHapticEffectType|, return [=a promise
+            rejected with=] reason {{TypeError}}.
+            </li>
+            <li>Let |document:Document?| be the [=current settings object=]'s 
+            [=relevant global object=]'s [=associated `Document`=].
+            </li>
+            <li>If |document| is `null` or |document| is not [=Document/fully
+            active=] or |document|'s [=Document/visibility state=] is
+            `"hidden"`, return [=a promise resolved with=]
+            {{GamepadHapticsResult/"preempted"}}.
```
marcoscaceres: These still look like exceptional situations to me... we should be returning [=a promise rejected with=] with something here.

extensions.html
marcoscaceres: That doesn't sound great because it seems to deviate away from playEffect()'s behavior. If they are using the same code path, then we should really do playEffect("pulse") as a followup.

index.html
marcoscaceres: I don't think we should be allowed to even get into the situation where we need a validator (i.e., we don't ever need to run custom validators for each value)... Consider, by merely changing the the WebIDL value for `startDelay` from double to an `unsigned long long` with a default of 0, we've already solved for the validity check without needing any prose (and it already gives you a TypeError for free). 

**I think we can get rid of "valid effect" entirely.**

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+                  </ol>
+                </li>
+              </ol>
+            </li>
+            <li>Return {{GamepadHapticActuator/[[playingEffectPromise]]}}.
+            </li>
+          </ol>
+        </dd>
+        <dt>
+          <dfn>reset()</dfn>
```
```suggestion
          <dfn>reset()</dfn> method
```

index.html
marcoscaceres: Ah! Ok, then maybe: 

```suggestion
            fixed-duration, constant-intensity vibration effect intended for an
```

index.html
marcoscaceres: Ok, so the only thing to check is if `weakMagnitude` and `strongMagnitude` are less than or equal to 1 (because it's the only thing WebIDL can't do today).

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+          <p>
+            The current effect was stopped or replaced (i.e., "preempted") by
+            another effect.
+          </p>
+        </dd>
+      </dl>
+    </section>
+    <section>
+      <h3>
+        <dfn>GamepadHapticEffectType</dfn> Enum
```
```suggestion
        <dfn>GamepadHapticEffectType</dfn> enum
```

index.html
```diff
@@ -1028,6 +1061,461 @@ <h2>
...
+          <dfn>startDelay</dfn> member
+        </dt>
+        <dd>
+          {{GamepadEffectParameters/startDelay}} sets the duration of the
+          delay after {{GamepadHapticActuator/playEffect()}} is called
+          until vibration is started, in milliseconds. During the delay
+          interval, the actuator must not vibrate. A <dfn>valid startDelay</dfn>
+          must be a positive value. An effect with an invalid
+          {{GamepadEffectParameters/startDelay}} is not a [=valid effect=]. If
+          the {{GamepadEffectParameters/startDelay}} is not specified, the
```
marcoscaceres: As above... just let WebIDL handler this through setting the data type to `unsigned long long`.

extensions.html
svillar: > > Ok, but if pulse() is Meta Quest only, that also puts it the "feature at risk". We should note that in the spec too, unless we expect to get a second implementation.
> > Is there a plan for a second implementation?
> 
> I'm fairly sure that this was also implemented in the Pico browser. @bkardell , was this added to Wolvic?

The version of Gecko that is shipped with the latests releases of wolvic (116) does implement Pulse()

marcoscaceres: Right, if we want to include it then we should do it as a followup with more detail about making it somewhat more interoperable (i.e., what does pulse actually do across various gamepads and how do we give them a consistent feel...). At the same time, we should add the behavior to be more controllable via vibrationActuator, because .pulse() is super crude.

index.html
nondebug: `unsigned long long` (milliseconds) should be fine for duration & startDelay, I don't think we need sub-millisecond timing for legacy rumble effects.

index.html
gabrielsanbrito: I have changed them to `unsigned long long`, but I think that `unsigned short` would be sufficient for this purpose, which would allow the actuator to vibrate for at most 65535 milliseconds (or ~65 seconds) (https://webidl.spec.whatwg.org/#idl-unsigned-short).

index.html
gabrielsanbrito: Great! I have removed this "MUST" part from here and from the "stop haptics effects" part. Also reworded it to be more cohesive.

index.html
gabrielsanbrito: Thanks! The magnitude values are being checked as part of the validation algorithm in the paragraph below.

index.html
gabrielsanbrito: Thanks. Moved it up to the `<dfn>GamepadEffectParameters</dfn> section`.

index.html
```diff
@@ -1028,6 +1061,454 @@ <h2>
...
+        <tr>
+          <td>
+            <dfn data-dfn-for=
+            "GamepadHapticActuator">[[\playingEffectPromise]]</dfn>
+          </td>
+          <td>
+            `null`
+          </td>
+          <td>
+            The {{Promise}} for the playing effect, or `null` if no effect is
```
```suggestion
            The {{Promise}} to play some effect, or `null` if no effect is
```

index.html
```diff
@@ -1028,6 +1061,454 @@ <h2>
...
+      </h2>
+      <p>
+        A {{GamepadHapticActuator}} corresponds to a configuration of motors or
+        other actuators that can apply a force for the purposes of haptic
+        feedback.
+      </p>
+      <pre class="idl">
+        [Exposed=Window]
+        interface GamepadHapticActuator {
+          readonly attribute FrozenArray&lt;GamepadHapticEffectType&gt; effects;
```
```suggestion
          [SameObject] readonly attribute FrozenArray&lt;GamepadHapticEffectType&gt; effects;
```

index.html
```diff
@@ -1028,6 +1061,454 @@ <h2>
...
+        </dt>
+        <dd>
+          <p>
+            Array of {{Gamepad/GamepadHapticEffectType}} values representing all
+            the types of haptic effects that the actuator supports. This
+            property MUST contain all the {{Gamepad/GamepadHapticEffectType}}
+            values that the actuator supports, unless the [=user agent=] does
+            not support playing effects of that type. The same object MUST be
+            returned and MUST NOT change for the lifetime of the
+            {{GamepadHapticActuator}}.
```
marcoscaceres: Moved the following to the IDL. 

```suggestion
            not support playing effects of that type.
```

index.html
```diff
@@ -1028,6 +1061,454 @@ <h2>
...
+      </table>
+      <dl data-dfn-for="GamepadHapticActuator">
+        <dt>
+          <dfn>effects</dfn> attribute
+        </dt>
+        <dd>
+          <p>
+            Array of {{Gamepad/GamepadHapticEffectType}} values representing all
+            the types of haptic effects that the actuator supports. This
+            property MUST contain all the {{Gamepad/GamepadHapticEffectType}}
```
```suggestion
            property lists the {{Gamepad/GamepadHapticEffectType}}
```

index.html
```diff
@@ -1028,6 +1061,454 @@ <h2>
...
+        <dd>
+          <p>
+            The {{GamepadHapticActuator/playEffect()}} method steps, called with
+            {{GamepadHapticEffectType}} |type:GamepadHapticEffectType| and
+            {{GamepadEffectParameters}} |params:GamepadEffectParameters |, are:
+          </p>
+          <ol class="algorithm">
+            <li>If |params:GamepadEffectParameters| does not describe a [=valid
+            effect=] of type |type:GamepadHapticEffectType|, return [=a promise
+            rejected with=] reason {{TypeError}}.
```
```suggestion
            rejected with=] a {{TypeError}}.
```

index.html
```diff
@@ -1028,6 +1061,454 @@ <h2>
...
+                <li>[=Issue a haptic effect=] to the actuator with |type| and
+                |params|.
+                </li>
+                <li>When the effect completes, if
+                [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}} is
+                not `null`, [=queue a global task=] on the [=relevant global
+                object=] of [=this=] using the [=gamepad task source=] to run
+                the following steps:
+                  <ol>
+                    <li>[=Resolve=]
```
marcoscaceres, line 1209:
```suggestion
                    <li>If  [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}} is `null`, abort these steps.</li>
                    <li>[=Resolve=]
```

index.html
```diff
@@ -1028,6 +1061,454 @@ <h2>
...
+        <ol>
+          <li>If
+          |actuator|.{{GamepadHapticActuator/[[playingEffectPromise]]}} is
+          `null`, abort these steps.
+          </li>
+          <li>[=Queue a global task=] on the [=relevant global object=] of
+          |actuator| using the [=gamepad task source=] to run the following
+          steps:
+            <ol>
+              <li>[=Resolve=]
```
marcoscaceres, line 1339:
```suggestion
              <li>If |actuator|.{{GamepadHapticActuator/[[playingEffectPromise]]}} is
          `null`, abort these steps.</li>
              <li>[=Resolve=]
```

index.html
```diff
@@ -1028,6 +1061,454 @@ <h2>
...
+        };
+      </pre>
+      <dl>
+        <dt>
+          <dfn>duration</dfn> member
+        </dt>
+        <dd>
+          {{GamepadEffectParameters/duration}} sets the duration of the
+          vibration effect in milliseconds. An effect with an invalid
+          {{GamepadEffectParameters/duration}} is not a [=valid effect=].
```
marcoscaceres: We don't need this, IMO:

```suggestion
          vibration effect in milliseconds.
```

index.html
```diff
@@ -1028,6 +1061,454 @@ <h2>
...
+        </dd>
+        <dt>
+          <dfn>startDelay</dfn> member
+        </dt>
+        <dd>
+          {{GamepadEffectParameters/startDelay}} sets the duration of the
+          delay after {{GamepadHapticActuator/playEffect()}} is called
+          until vibration is started, in milliseconds. During the delay
+          interval, the actuator must not vibrate. An effect with an invalid
+          {{GamepadEffectParameters/startDelay}} is not a [=valid effect=].
```
marcoscaceres: Nothing is guaranteed here. I'd suggest: 

```suggestion
          until vibration is started, in milliseconds. During the delay
          interval, the actuator SHOULD NOT vibrate.
```

marcoscaceres (approved):
A couple of nits and suggestions, but generally looking pretty solid now. 

Outstanding work @gabrielsanbrito. Thank you so much for putting so much work into this. 

@nondebug, do you want to give this one last look over?

marcoscaceres:
File Gecko and WebKit bugs... WebKit already implements vibratorActuator, and so does Chrome. But it would be good to make sure we are all interoperable. 

In any case, it meets criteria for inclusion in the spec.

index.html
gabrielsanbrito: Ok. I am editing it to return a promise rejected with an "InvalidStateError" DOMException both for `playEffect` and `reset`.

index.html
gabrielsanbrito: Will edit to return a promise rejected with an "{{InvalidStateError}}" {{DOMException}}.

marcoscaceres:
@gabrielsanbrito, I merged in my suggestions and rebased. 

The last remaining thing is just adding the DOMHighResTimestamp suggestion. Do you want me to do that?

gabrielsanbrito:
> @gabrielsanbrito, I merged in my suggestions and rebased.
> 
> The last remaining thing is just adding the DOMHighResTimestamp suggestion. Do you want me to do that?

Thanks for doing that, @marcoscaceres! I was looking up on how to do the DOMHighResTimestamp. I was thinking on something like this:
```
<li>Let {{GamepadHapticActuator/[[playingEffectPromise]]}} be [=a
new promise=].
<li>Let |playEffectTimestamp:DOMHighResTimestamp| be the [=current high resolution time=] given the |document|'s [=relevant global object=].
</li>
<li>Do the following steps [=in parallel=]:
  <ol>
    <li>[=Issue a haptic effect=] to the actuator with |type|, |params|, and the 
     |playEffectTimestamp|.
    </li>
[...]
```

Then, on the <dfn>issue a haptic effect</dfn>, I was thinking of adding:
```
To <dfn>issue a haptic effect</dfn> on an actuator, the [=user agent=] MUST send a command to the device to render an effect of |type:GamepadHapticEffectType| and try to make it use the provided |params:GamepadEffectParameters|. The [=user agent=] SHOULD use the provided |playEffectTimestamp:DOMHighResTimestamp| for more precise playback timing when |params|.{{startDelay}} is not `0.0`.
```

What do you think? Please feel free to add it directly if it is too far from ideal.

gabrielsanbrito:
> File Gecko and WebKit bugs... WebKit already implements vibratorActuator, and so does Chrome. But it would be good to make sure we are all interoperable.
> 
> In any case, it meets criteria for inclusion in the spec.

It looks like both WebKit (https://bugs.webkit.org/show_bug.cgi?id=267523) and Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=1824217) already have bugs filed for the vibrationActuator addition. I have just added a small update comment on the Gecko one to let them know that `vibrationActuator` is being added to the Gamepad API spec.

marcoscaceres:
@gabrielsanbrito, the proposed text for the timestamp sounds great!

marcoscaceres:
Ok, cool. I've updated the OP to link to the existing Gecko bug. Thanks also for commenting there @gabrielsanbrito.

marcoscaceres:
I wonder if we should do a little "test fest" on our next call. Would be nice to just run through the algorithms manually and check for any introp issues.

toji (approved):
Change LGTM, thanks for the update! And, to reiterate, I agree with the decision to leave `pulse()` in the extensions doc, given the confusion around it and it's intended behavior.

I appreciate the explicit "completed"/"preempted" status of `playEffect()` in this regard, and would recommend that if `pulse()` is carried forward that it may be worth evaluating whether or not to transition to that model for consistency and clarity. (Since the enums will evaluate to be truthy and existing implementations have apparently only ever returned true this is maybe a reasonable backwards compat break.)

gabrielsanbrito:
@marcoscaceres had to remove the `"switch"` class from the `dl` tag. It was messing the numbers of the ordered list. With the class set, I was having:
**Item 1**: Given the value of {{GamepadHapticEffectType}} |type:GamepadHapticEffectType|, switch on:
&emsp;`"dual-rumble"`: If params does not describe a [valid dual-rumble effect](https://github.com/w3c/gamepad/pull/190#dfn-valid-dual-rumble-effect), return false.
**Item 3**: Return `true`.

Just removing the class made "Return `true`." become item 2 again - so I have removed it. I don't know if I am missing on something here or if there is something wrong with the `"switch"` class. Please let me know if this looks good for you. Now it is looking like this:
**Item 1**: Given the value of {{GamepadHapticEffectType}} |type:GamepadHapticEffectType|, switch on:
&emsp;`"dual-rumble"`: If params does not describe a [valid dual-rumble effect](https://github.com/w3c/gamepad/pull/190#dfn-valid-dual-rumble-effect), return false.
**Item 2**: Return `true`.

marcoscaceres:
Yeah, that's fine. And don't worry, we will eventually get rid of all the `<dl>`s in algorithms... they are annoying.

marcoscaceres:
@gabrielsanbrito, just waiting on Matt. Spoke to him and he's hoping to do the final review soon. Expecting this to land hopefully early next week.

nondebug (approved):
lgtm

gabrielsanbrito:
Thank you all for the valuable feedback! 

@marcoscaceres may I merge the pull request?

marcoscaceres:
Merged 🎉

razvanphp:
This link in the OP is missing: https://issues.chromium.org/issues/40834175

PR #193, opened 2024-02-09
Allow initializing the gamepad index to a specific value

nondebug:
Closes #???

Implementation commitment:

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/pull/193.html" title="Last updated on Feb 9, 2024, 7:09 PM UTC (be68590)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/193/4b33578...be68590.html" title="Last updated on Feb 9, 2024, 7:09 PM UTC (be68590)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -650,8 +650,9 @@
           Constructing a `Gamepad`
         </h3>
         <p>
-          <dfn>A new `Gamepad`</dfn> representing a connected gamepad device is
-          constructed by performing the following steps:
+          <dfn>A new `Gamepad`</dfn> with optional |index:long| representing a
+          connected gamepad device is constructed by performing the following
+          steps:
         </p>
         <ol>
           <li>Let |gamepad:Gamepad| be a newly created {{Gamepad}} instance:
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -650,17 +650,24 @@ <h3>
           Constructing a `Gamepad`
         </h3>
         <p>
-          <dfn>A new `Gamepad`</dfn> representing a connected gamepad device is
-          constructed by performing the following steps:
```
marcoscaceres, line 654: Who provides the index originally? I guess it's the UA. I wonder if we should make this more consistent by making it mandatory to pass the index. Would that work across browsers?

nondebug: The UA shouldn't decide the index, it's decided by the "select an unused gamepad index" algorithm which is based on the sequence of previous gamepad connections/disconnections.

Let's make |index| mandatory by moving index selection into the "when a gamepad becomes available on the system" steps.

nondebug: Looking at this again I now remember why it's done this way. The "select an unused gamepad index" algorithm needs to access the Navigator object. Currently it does so by creating a Gamepad object and then finding the navigator via the Gamepad's relevant global object. If we want to pass the index into the "a new Gamepad with index" steps then we need to get the Navigator object from somewhere else, since the Gamepad doesn't exist yet. This would need to happen in the "When a gamepad becomes available on the system" steps.

Is there a way to find the relevant global object without referencing an existing object? The current method seems kind of magical, how did the Gamepad know what its relevant global object should be when it was created? We could abuse the same trick by creating a dummy object just to get its global object, but I assume there's a better way to specify this that doesn't involve creating a dummy object.

PR #196, opened 2024-03-05
Add extended buttons to support trackpad (#191)

### LINKED ISSUE #191: Add trackpad button support
The current W3C Gamepad specification does not include mappings for the trackpad button on Sony DualShock and DualSense controllers. This omission leads to inconsistencies across browsers, as observed in WebKit (Safari) and Chromium (Chrome).

### Suggested Update:
1. **Standardize Trackpad Button Mapping:** Introduce a standard mapping for the trackpad button, possibly as button 17, to align with Chromium's existing implementation.
2. **Uniform Button Array Size:** Ensure a 17-element button array for all gamepad controllers to maintain consistency and aid in fingerprinting mitigation.

### Additional Information:
This inconsistency was noted in WebKit bug report [266293](https://bugs.webkit.org/show_bug.cgi?id=266293), with a test conducted on gamepad-tester.com showing different behaviors in Safari Technology Preview and Chrome.

### Impact:
Addressing this will enhance cross-browser compatibility and user experience for gamepad users, especially those using Sony's controllers.

**References:**
- WebKit Bug Report: [266293 – Gamepad: Map DualShock/DualSense Trackpad Button](https://bugs.webkit.org/show_bug.cgi?id=266293)

xingri:
Closes #191 

Implementation commitment:

 * [ ] WebKit (https://bugs.webkit.org/show_bug.cgi?id=266293)
 * [ ] Chromium (https://issues.chromium.org/issues/339841686)

***
<a href="https://pr-preview.s3.amazonaws.com/xingri/gamepad/pull/196.html" title="Last updated on Jul 10, 2025, 11:10 PM UTC (eeea475)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/196/c430e73...xingri:eeea475.html" title="Last updated on Jul 10, 2025, 11:10 PM UTC (eeea475)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -820,8 +820,11 @@
           <li>[=list/For each=] |rawInputIndex:long| of [=the range=] from 0 to
           |inputCount| − 1:
             <ol>
-              <li>If the the gamepad button at index |rawInputIndex|
-              [=represents a Standard Gamepad button=]:
+              <li>Let |type| be the result of determining if the button at index |rawInputIndex|
+              [=represents a Standard Gamepad button=] or represents a [=additional gamepad button=].
+              If it's neither, use {{GamepadButtonType/"non-standard"}}.
+              </li>
+              <li>If |type| is not {{GamepadButtonType/"non-standard"}}:
                 <ol>
                   <li>Let |canonicalIndex:long| be the [=canonical index=] for
                   the button.
@@ -878,8 +881,17 @@
           <li>Initialize |buttons| to be an empty [=list=].
           </li>
           <li>[=list/For each=] |buttonIndex:long| of [=the range=] from 0 to
-          |buttonsSize| − 1, [=list/append=] a [=new=] {{GamepadButton}} to
-          |buttons|.
+          |buttonsSize| − 1:
+            <ol>
+              <li>Let |button:GamepadButton| be a [=new=] {{GamepadButton}} with its
+              |button|.{{GamepadButton/pressed}} initialized to `false`,
+              |button|.{{GamepadButton/touched}} initialized to `false`,
+              |button|.{{GamepadButton/type}} initialized to |type|, and
+              |button|.{{GamepadButton/value}} initialized to 0.
+              </li>
+              <li>[=list/Append=] |button| to |buttons|.
+              </li>
+            </ol>
           </li>
           <li>Return |buttons|.
           </li>
@@ -900,6 +912,7 @@
           readonly attribute boolean pressed;
           readonly attribute boolean touched;
           readonly attribute double value;
+          readonly attribute GamepadButtonType type;
         };
       </pre>
       <p>
@@ -1017,6 +1030,13 @@
             <li>Return [=this=].{{GamepadButton/[[value]]}}.
             </li>
           </ol>
+        </dd>
+        <dt>
+          <dfn>type</dfn> attribute
+        </dt>
+        <dd>
+          An enumerated {{GamepadButtonType}} attribute that classifies the current button's type in relation to an
+          extended mapping.
         </dd>
       </dl>
     </section>
@@ -2031,6 +2051,60 @@
           Visual representation of a [=Standard Gamepad=] layout.
         </figcaption>
       </figure>
+      <section>
+        <h3>
+          <dfn data-lt="additional gamepad button">Additional gamepad buttons</dfn>
+        </h3>
+        <p>
+          This section introduces an extended gamepad button mapping beyond the [=Standard Gamepad=] mapping.
+          These additional buttons are commonly found on certain gamepad models.
+          Some examples of these additional buttons include trackpads or touchpads, share or capture buttons,
+          voice assistant buttons, home buttons, and various squeeze buttons.
+          It's important to note that this list is not exhaustive, and user agents may utilize different
+          or additional buttons for these or other gamepad models.
+          Consequently, the number of buttons on the {{Gamepad}} is not limited to the standard mapping of 17 buttons.
+        </p>
+        <section>
+          <h4>
+            <dfn>GamepadButtonType</dfn> Enum
+          </h4>
+          <p>
+            To accommodate additional gamepad buttons, we have defined an enumeration for the various button types termed
+	    {{GamepadButtonType}}, and have expanded the {{GamepadButton}} interface to encompass this new
+	    {{GamepadButtonType}} enumeration.
+          </p>
+          <p>
+            This enum defines the set of possible button types.
+          </p>
+          <pre class="idl">
+            enum GamepadButtonType {
+              "standard",
+              "non-standard",
+              "trackpad",
+            };
+          </pre>
+          <dl data-dfn-for="GamepadButtonType">
+            <dt>
+              "<dfn>standard</dfn>"
+            </dt>
+            <dd>
+              Represent a button has a button type defined in the [=Standard Gamepad=] mapping.
+            </dd>
+            <dt>
+              "<dfn>non-standard</dfn>"
+            </dt>
+            <dd>
+              Represents a button that exists but doesn't have a standard name.
+            </dd>
+            <dt>
+              "<dfn>trackpad</dfn>"
+            </dt>
+            <dd>
+              Represent a trackpad input type.
+            </dd>
+          </dl>
+        </section>
+      </section>
       <section>
         <h3>
           Fingerprinting mitigation
--- a/index.html
+++ b/index.html
@@ -2082,8 +2082,8 @@
           </p>
           <pre class="idl">
             enum GamepadButtonType {
-              "standard",
               "non-standard",
+              "standard",
               "trackpad",
             };
           </pre>
--- a/index.html
+++ b/index.html
@@ -2057,7 +2057,7 @@
       </figure>
       <section>
         <h3>
-          <dfn data-lt="additional gamepad button">Additional gamepad buttons</dfn>
+          <dfn>Additional gamepad buttons</dfn>
         </h3>
         <p>
           This section introduces an extended gamepad button mapping beyond the [=Standard Gamepad=] mapping.


extensions.html
```diff
@@ -649,6 +649,168 @@ <h2>
         </p>
       </dd>
     </section>
+    <section>
```
nondebug: Let's define extended buttons in the main spec document (index.html). The extensions document is not officially part of the spec and we want to eventually remove it.

The spec PR should extend [GamepadButton](https://www.w3.org/TR/gamepad/#gamepadbutton-interface) to add a new attribute (perhaps called `GamepadButton.name`) and an algorithm for the getter. The type of the new attribute will be a new enum (perhaps called `GamepadButtonName`). The enum type should be defined in the spec along with all its values.

Add steps to initialize the new attribute when [initializing buttons](https://www.w3.org/TR/gamepad/#dfn-initializing-buttons). The steps should define what to do if a gamepad is missing a button defined in the Standard Gamepad. I think it makes sense to use a special value for these buttons so applications can identify which buttons are not actually mapped. (Ideally we'd set the button to null, but this would break existing apps which don't check for null.)

Also, initialization steps should define what to do if a gamepad has a button that doesn't have a predefined Standard Gamepad index or enum value. Do we enumerate it? If so, what should the new attribute's value be? Perhaps we can add a special "unrecognized" value for this.

extensions.html
```diff
@@ -649,6 +649,168 @@ <h2>
...
+      <h2>
+        "extended" Gamepad Buttons Mapping
+      </h2>
+      <p>
+        This section defines extended gamepad buttons mapping beyonds the Standard Gamepads layout,
+        but are commonly found on some gamepad models .The <code>_ext</code> suffix is used to indicate
```
nondebug: The `_ext` suffix comes from OpenXR's input subpaths specification:

https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#_adding_input_sources_via_extensions

We should decide whether this attribute uses OpenXR input subpaths or something new that we want to define for ourselves. If we're using OpenXR subpaths then we should explicitly link to the OpenXR spec which explains how to extend it. If we're only using it for inspiration then I think we should drop the `_ext` suffix.

extensions.html
```diff
@@ -649,6 +649,168 @@ <h2>
...
+            <td>dpad_right</td>
+            <td>Standard: Right button in left cluster</td>
+          </tr>
+          <tr>
+            <td>buttons[16]</td>
+            <td>system</td>
+            <td>Standard: Center button in center cluster</td>
+          </tr>
+          <tr>
+            <td>buttons[17]</td>
```
nondebug: This suggests the "trackpad" button can only appear at `buttons[17]`. My thinking is that we should not continue the current pattern of assigning meaning to button indices, otherwise the `buttons` array will be very sparse since most gamepads do not share the same extra buttons. Instead of using button indices to identify buttons, apps can check the new enum attribute.

For this non-normative section, I think it will be clearer if we provide a separate table for each gamepad instead of trying to combine them all into a single table.

xingri: @nondebug  Thanks for the review. Will update the PR based on the feedback shortly.

xingri: @nondebug I have updated the PR based on your feedback. I have added the enum for identifying the gamepad buttons. Please review this changes and share your opinions.

index.html
```diff
@@ -1718,232 +1718,429 @@ <h2>
...
+            This enum defines the Extended type of Gamepads.
+          </p>
+          <pre class="idl">
+            enum partial-GamepadMappingType {
+              "extended"
+            };
+          </pre>
+          <dl>
+            <dt>
+              <dfn>`"extended"`</dfn>
```
nondebug: Introducing a new value for GamepadMappingType will break applications that currently check for "standard". I don't think it's necessary to add a new enum value for this, applications can discover whether the gamepad supports extra buttons by inspecting the buttons array.

index.html
```diff
@@ -1718,232 +1718,429 @@ <h2>
...
+        <section>
+          <h2>
+            Partial <dfn>GamepadButton</dfn> Interface
+          </h2>
+          <p>
+            This partial interface extends the GamepadButton interface in this spec.
+          </p>
+          <pre class="idl">
+            partial interface GamepadButton {
+              readonly attribute GamepadButtonType type;
```
nondebug: Please move the definition to the main `interface GamepadButton` IDL section.

index.html
```diff
@@ -1718,232 +1718,429 @@ <h2>
...
+              <th>
+                Initial value
+              </th>
+              <th>
+                Description (non-normative)
+              </th>
+            </tr>
+            <tr>
+              <td>
+                <dfn data-dfn-for="GamepadButton">[[\type]]</dfn>
```
nondebug: I don't think we need an internal slot.

For other attributes we use a slot because the attribute is read-only but the implementation needs to be able to change the value. In this case the value should never change after initialization.

index.html
```diff
@@ -1718,232 +1718,429 @@ <h2>
...
+                `undefined`
+              </td>
+              <td>
+                Indicates the button type the controller is held in.
+              </td>
+            </tr>
+	  </table>
+          <dl data-dfn-for="GamepadButton">
+          <dt>
+            <dfn>type</dfn>
```
nondebug: `<dfn>type</dfn> attribute`

index.html
```diff
@@ -1718,232 +1718,429 @@ <h2>
...
+              The Gamepad's controls have been mapped to the Extended Gamepad layout.
+            </dd>
+          </dl>
+        </section>
+	<section>
+          <h3>
+            <dfn>GamepadButtonType</dfn> Enum
+          </h3>
+          <p>
+            This enum defines the set of possible botton types.
```
nondebug: botton -> button

index.html
```diff
@@ -1718,232 +1718,429 @@ <h2>
...
+          </p>
+          <pre class="idl">
+            enum GamepadButtonType {
+              "",  /* unknown, or not applicable */
+              "trackpad",
+              "share",
+              "..."
```
nondebug: Omit "..." since it is not a valid enum value.

sony_dualshock4_gamepad.svg
nondebug, line 1: The diagrams are difficult to read. For clarity I think we should provide this information as a table instead of a diagram.

index.html
```diff
@@ -1718,232 +1718,429 @@ <h2>
...
+                Indicates the button type the controller is held in.
+              </td>
+            </tr>
+	  </table>
+          <dl data-dfn-for="GamepadButton">
+          <dt>
+            <dfn>type</dfn>
+          </dt>
+          <dd>
+            An enumeration, {{GamepadButtonType}}, that indicates which button types the controller
```
nondebug: The description for the `type` attribute needs to be updated

xingri: @nondebug  Again thank you for the review. I have no objections and will update this PR based on your feedbacks.

index.html
```diff
@@ -952,6 +953,17 @@ <h2>
...
         </tr>
+        <tr>
+          <td>
+            <dfn data-dfn-for="GamepadButton">[[\type]]</dfn>
+          </td>
+          <td>
+            `undefined`
+          </td>
+          <td>
+            Indicates the button type the controller is held in.
```
```suggestion
            The button type of the controller.
```

index.html
```diff
@@ -1718,232 +1737,341 @@ <h2>
...
+	<section>
+          <h3>
+            <dfn>GamepadButtonType</dfn> Enum
+          </h3>
+          <p>
+            This enum defines the set of possible button types.
+          </p>
+          <pre class="idl">
+            enum GamepadButtonType {
+              "",  /* unknown, or not applicable */
```
marcoscaceres: Better to make it explicit.

```suggestion
              "unknown",
```

And if not applicable, maybe we should make . type nullable?

index.html
```diff
@@ -1718,232 +1737,341 @@ <h2>
...
-          </tr>
-        </tbody>
-      </table>
-      <figure>
-        <img src="standard_gamepad.svg" alt="">
-        <figcaption>
-          Visual representation of a [=Standard Gamepad=] layout.
-        </figcaption>
-      </figure>
+      <section>
```
marcoscaceres, line 2470: @xingri, would it be possible to revert some of this so it only shows what you changed?

xingri: @marcoscaceres  thanks for the review. updated the changes by your feedback.

index.html
```diff
@@ -1018,6 +1030,13 @@ <h2>
             </li>
           </ol>
         </dd>
+        <dt>
+          <dfn>type</dfn> attribute
+        </dt>
+        <dd>
+          An enumeration, {{GamepadButtonType}}, that indicates which button types the controller
+          is being held in.
+        </dd>
```
mrmcpowned, line 1397: Looks like we're still missing getter definitions, no?

index.html
```diff
@@ -1018,6 +1030,13 @@ <h2>
             </li>
           </ol>
         </dd>
+        <dt>
+          <dfn>type</dfn> attribute
+        </dt>
+        <dd>
+          An enumeration, {{GamepadButtonType}}, that indicates which button types the controller
+          is being held in.
```
mrmcpowned: Is this phrasing in line with other specs? There's an [open spec issue](https://github.com/whatwg/html/issues/9832) regarding consistency in enum phrasing which makes me think this would be better phrased as:
> An enumerated {{GamepadButtonType}} attribute that classifies the current button's type in relation to an extended mapping

index.html
```diff
@@ -1944,6 +1963,109 @@ <h2>
...
+          into the GamepadMappingType. Additionally, we have defined an enumeration for the various
+          button types in the table, termed "GamepadButtonType", and have expanded the "GamepadButton"
+          interface to encompass this new button type enumeration.
+       </p>
+        <table class="tg">
+          <thead>
+            <tr>
+              <th scope="col">Buttons</th>
+              <th scope="col">Gamepad Models</th>
+            </tr>
```
mrmcpowned: It feels like there's some disconnect with this table and the available GamepadButtonType enums. Are we missing a possible third column which maps each of these button to a respective GamepadButtonType?

index.html
```diff
@@ -1944,6 +1963,109 @@ <h2>
...
+            <tr>
+              <td>squeeze_right_lower</td>
+              <td>Xbox Elite Wireless Controller</td>
+            </tr>
+          </tbody>
+        </table>
+       <section>
+          <h3>
+            <dfn>GamepadButtonType</dfn> Enum
```
mrmcpowned, line 2485: I see 8 different classes of extended buttons in the table above, but only 2 applicable GamepadButtonTypes in the enum definition. Is the intention to expand the enum list below so we have a 1:1 mapping with the rows in the table?

index.html
```diff
@@ -1944,6 +1963,109 @@ <h2>
...
+            </tr>
+           <tr>
+              <td>voice assistant</td>
+              <td>Google Stadia Controller<br>
+                Amazon Luna Controller<br>
+              </td>
+            </tr>
+            <tr>
+              <td>home</td>
+              <td>Nvidia Shield Gamepad</td>
```
mrmcpowned: The Xbox and PlayStation controllers have their own respective home buttons. Are we intending to add those to this list or is there some special reason why we're classifying this separately for the Shield gamepad?

index.html
```diff
@@ -1944,6 +1963,109 @@ <h2>
...
+              "trackpad",
+              "share"
+            };
+          </pre>
+          <dl data-dfn-for="GamepadButtonType">
+            <dt>
+              <dfn>unknown</dfn>
+            </dt>
+            <dd>
+              This indicates button type is unknown or not applicable.
```
mrmcpowned: Is there a tangible benefit for User Agents to populate all the GamepadButtons with `unknown` for any buttons that can't be classified as one of the other types? Would it be better to simply have the attribute as `undefined` when that's the case?

xingri: @mrmcpowned  Thanks for the review. I agree with your feedbacks. Will update this changes shortly.

index.html
xingri: updated.

index.html
xingri: updated.

index.html
xingri: Thanks. updated.

index.html
xingri: Agree with your feedback @mrmcpowned, updated.

index.html
```diff
@@ -1944,6 +1963,132 @@ <h2>
...
+          <thead>
+            <tr>
+              <th scope="col">Buttons</th>
+              <th scope="col">Gamepad Models</th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr>
+              <td>trackpad/touchpad</td>
+              <td>Sony DualSense, DualShock 4</td>
```
marcoscaceres: As I imagine these products will change a lot over time, so these should probably be in [Wiki page](https://github.com/w3c/gamepad/wiki) (or if we want to formalize it, and actual registry). We can point the spec to the wiki page.

index.html
```diff
@@ -1944,6 +1963,132 @@ <h2>
...
+              "squeeze_right",
+              "squeeze_right_lower"
+            };
+          </pre>
+          <dl data-dfn-for="GamepadButtonType">
+            <dt>
+              <dfn>trackpad</dfn>
+            </dt>
+            <dd>
+              This button is being assigned to trackpad button.
```
marcoscaceres: Probably just ok to say:
 
```suggestion
              A trackpad button.
```

Same for the rest.

index.html
```diff
@@ -1944,6 +1963,132 @@ <h2>
...
       </figure>
+      <section>
+        <h3>
+          Extended Mapping
+        </h3>
+        <p>
+          This section introduces an extended gamepad button mapping beyond the Standard Gamepad layout.
+          These additional buttons are commonly found on certain gamepad models.
+          The following table defines the extended buttons used by some gamepad models in the following diagrams.
+          It’s important to note that this list is not exhaustive, and user agents may utilize different
```
```suggestion
          It's important to note that this list is not exhaustive, and user agents may utilize different
```

index.html
```diff
@@ -1944,6 +1963,132 @@ <h2>
...
+          or additional buttons for these or other gamepad models.
+          Consequently, the number of buttons on the Gamepad is not limited to the standard mapping of 17 buttons.
```
```suggestion
          Consequently, the number of buttons on the {{Gamepad}} is not limited to the standard mapping of 17 buttons.
```

index.html
```diff
@@ -1944,6 +1963,132 @@ <h2>
...
+        </p>
+       <p>
+          To accommodate extended gamepads, we have incorporated an "extended" mapping enumeration
+          into the GamepadMappingType. Additionally, we have defined an enumeration for the various
```
```suggestion
          into the {{GamepadMappingType}}. Additionally, we have defined an enumeration for the various
```

index.html
```diff
@@ -1944,6 +1963,132 @@ <h2>
...
+          button types in the table, termed "GamepadButtonType", and have expanded the "GamepadButton"
```
```suggestion
          button types in the table, termed {{GamepadButtonType}}, and have expanded the {{GamepadButton}}
```

index.html
```diff
@@ -1944,6 +1963,132 @@ <h2>
...
+       <section>
+          <h3>
+            <dfn>GamepadButtonType</dfn> Enum
+          </h3>
+          <p>
+            This enum defines the set of possible button types.
+          </p>
+          <pre class="idl">
+            enum GamepadButtonType {
+              "trackpad",
```
marcoscaceres, line 2499: Nit: Please alpha sort these

index.html
xingri: @marcoscaceres  Thanks for the feedback. I have updated the rest of the comments.
I agree with you that it will keep changing and it would be great to have the link inside of the gamepad working group wiki. If this is not a hard requirement, I would like to follow up creating the wiki page on the different button mappings of the game controllers mentioned in this table as a separate PR.

index.html
```diff
@@ -952,6 +953,17 @@ <h2>
             .. 1.0]
           </td>
         </tr>
+        <tr>
+          <td>
+            <dfn data-dfn-for="GamepadButton">[[\type]]</dfn>
```
nondebug: I think the internal slot [[type]] isn't needed and can be removed.

Normally we use internal slots for read-only attributes that the user agent needs to modify. We expect a GamepadButton created with a button type will always have that type, so we don't need a way to modify it.

index.html
nondebug: sgtm, let's remove the table from this PR and follow up with the wiki page.

index.html
nondebug, line 1: Please add steps to the "initialize buttons" steps to initialize the button type.

https://w3c.github.io/gamepad/#dfn-initializing-buttons

Currently we initialize buttons in two phases. First, determine how many buttons are needed. Then, create all the buttons.

We'll need to remember the button type from the first phase so we can initialize the GamepadButton in the second phase. We also need to handle some special cases:

A GamepadButton may be created without a corresponding button input on the gamepad. (For instance, a gamepad may not have all the Standard Gamepad buttons; the button array entry for a missing button will still have a GamepadButton object but its value should never change.) We can initialize GamepadButton.type to null to represent this case.

A gamepad may have buttons that don't have a canonical button type. We should create a special GamepadButtonType enum value to signal that the button exists but doesn't have a canonical name yet.

xingri: @nondebug  Thanks for the feedback. Added steps for considering the button types.

index.html
```diff
@@ -2031,6 +2051,62 @@ <h2>
...
         </figcaption>
       </figure>
+      <section data-dfn-for="ExtendedMapping" data-link-for="ExtendedMapping">
+        <h3>
+	  <dfn>ExtendedMapping</dfn>
+        </h3>
+        <p>
+          This section introduces an extended gamepad button mapping beyond the Standard Gamepad layout.
+          These additional buttons are commonly found on certain gamepad models.
+          The following table defines the extended buttons used by some gamepad models.
```
```suggestion
          The following definition list defines the <dfn>extended gamepad buttons</dfn> used by some gamepad models.
```

index.html
```diff
@@ -820,8 +820,13 @@ <h3>
           <li>[=list/For each=] |rawInputIndex:long| of [=the range=] from 0 to
           |inputCount| − 1:
             <ol>
-              <li>If the the gamepad button at index |rawInputIndex|
-              [=represents a Standard Gamepad button=]:
+              <li>Determine the type of the button at index |rawInputIndex|
+              [=represents a Standard Gamepad button=] or Extended button
+              types as described in the {{ExtendedMapping}} using the provided
+              buttonTypes array. If the button does not have a canonical type,
+              use GamePadButtonType.non_canonical.
```
marcoscaceres: This should be specified as: 

```suggestion
              <li>Let |button| be the button at index |rawInputIndex|.</li>
              <li>Let |type| be the result of determining if the |button|'s type represents a [=Standard Gamepad button|represents a Standard Gamepad button=] or an [=Extended Gamepad button|represents a Extended Gamepad button=]. If it's neither, use {{GamePadButtonType/"non_canonical"}}. 
              </li>
```

And `<dfn>represents a Standard Gamepad button</dfn>` should be defined underneath where "represents a Standard Gamepad button" is defined/

index.html
```diff
@@ -820,8 +820,13 @@ <h3>
...
+              </li>
+              <li>If the button type is not GamepadButtonType.non_canonical:
```
```suggestion
              <li>If |type| is not {{GamepadButtonType/"non_canonical"}}:
```

index.html
```diff
@@ -878,8 +883,15 @@ <h3>
           <li>Initialize |buttons| to be an empty [=list=].
           </li>
           <li>[=list/For each=] |buttonIndex:long| of [=the range=] from 0 to
-          |buttonsSize| − 1, [=list/append=] a [=new=] {{GamepadButton}} to
-          |buttons|.
+          |buttonsSize| − 1:
+            <ol>
+              <li>Determine the type of the button from the provided buttonTypes
```
marcoscaceres: This should copy the steps above, with the correct linking.

index.html
marcoscaceres: Yeah, though it's implied... but yeah, if the others have the getter steps (what it was initialized to) then so should this.

index.html
```diff
@@ -2031,6 +2051,62 @@ <h2>
           Visual representation of a [=Standard Gamepad=] layout.
         </figcaption>
       </figure>
+      <section data-dfn-for="ExtendedMapping" data-link-for="ExtendedMapping">
+        <h3>
+	  <dfn>ExtendedMapping</dfn>
+        </h3>
+        <p>
```
marcoscaceres: Oh wait. ExtendedMapping is not an IDL thing... it's a concept. 

```suggestion
      <section>
        <h3>
	  <dfn>Extended Gamepad Mapping</dfn>
        </h3>
        <p>
```

index.html
```diff
@@ -2031,6 +2051,62 @@ <h2>
...
+          This section introduces an extended gamepad button mapping beyond the Standard Gamepad layout.
```
```suggestion
          This section introduces an extended gamepad button mapping beyond the [=Standard Gamepad mapping=].
```

index.html
```diff
@@ -2031,6 +2051,62 @@ <h2>
...
+          These additional buttons are commonly found on certain gamepad models.
+          The following table defines the extended buttons used by some gamepad models.
+          It's important to note that this list is not exhaustive, and user agents may utilize different
+          or additional buttons for these or other gamepad models.
+          Consequently, the number of buttons on the {{Gamepad}} is not limited to the standard mapping of 17 buttons.
+        </p>
+       <p>
+          To accommodate extended gamepads, we have incorporated an "extended" mapping enumeration
```
marcoscaceres: It seems the "extended" value hasn't been added to GamepadMappingType yet. Can you please update the PR and add it,  along with this definition?

index.html
```diff
@@ -2031,6 +2051,62 @@ <h2>
...
+          </h3>
+          <p>
+            This enum defines the set of possible button types.
+          </p>
+          <pre class="idl">
+            enum GamepadButtonType {
+              "standard",
+              "non-canonical",
+               "trackpad", 
+               // ... More enum types to be defined by the separate W3C wiki page.
```
marcoscaceres: We can do this, but we need to change GamepadButtonType to an DOMString. Otherwise, the enum will always be out of date. Alternatively, we don't use the Wiki and just use the spec... it might be better to just use the spec, so we can have more oversight over what is defined here (and get better interop).

index.html
```diff
@@ -2031,6 +2051,62 @@ <h2>
...
+            };
+          </pre>
+          <dl data-dfn-for="GamepadButtonType">
+            <dt>
+              <dfn>standard</dfn>
+            </dt>
+            <dd>
+              Represent a button has a button type defined in the standard mapping.
```
```suggestion
              Represent a button has a button type defined in the [=standard gamepad mapping=].
```

index.html
```diff
@@ -2031,6 +2051,62 @@ <h2>
...
+            </dd>
+            <dt>
+              <dfn>non-canonical</dfn>
+            </dt>
+            <dd>
+              Represents a button that exists but doesn't have a canonical name.
+            </dd>
```
marcoscaceres: Maybe clearer:

```suggestion
            <dt>
              <dfn>non-standard</dfn>
            </dt>
            <dd>
              Represents a button that exists but doesn't have a standardized name.
            </dd>
```

index.html
```diff
@@ -2031,6 +2051,62 @@ <h2>
...
+            <dt>
+              <dfn>trackpad</dfn>
+            </dt>
+            <dd>
+              Represent a trackpad input type which beyonds the scope of the standard mapping.
```
```suggestion
              Represent a trackpad input type.
```

index.html
```diff
@@ -2031,6 +2051,62 @@ <h2>
...
+	      <p>More enum types to be defined by the separate W3C wiki page.
```
marcoscaceres: On reflection, we should do this in the spec... 

```suggestion
```

marcoscaceres (changes requested):
Some suggested changes.

xingri:
> Some suggested changes.

@marcoscaceres thank you so much for the detailed feedbacks.
Will prepare the revised version based on your feedback soon.

xingri:
> Some suggested changes.

@marcoscaceres Thanks for the feedback. Applied most of the change requests. 

Regarding the extended mapping, based on the working group discussion on last March with @nondebug while you were away, 
we discussed that we will not have the additional mapping type of "extended" for those gamepads that have a mapping that is "standard" plus extended bits.  So, we'd keep "standard" but just have more controls.

Let's discuss a little bit during the call. Also I will fix the syntax error soon.

xingri:
@marcoscaceres, @nondebug  Please review the updated sentences.
Also linked the webkit & chromium issues.

index.html
```diff
@@ -820,8 +820,13 @@ <h3>
           <li>[=list/For each=] |rawInputIndex:long| of [=the range=] from 0 to
           |inputCount| − 1:
             <ol>
-              <li>If the the gamepad button at index |rawInputIndex|
-              [=represents a Standard Gamepad button=]:
+              <li>Let |button| be the button at index |rawInputIndex|.
+              </li>
+              <li>Let |type| be the result of determining if the |button| [=represents
+              a Standard Gamepad button=] or represents [=Additional gamepad buttons=]
```
nondebug: Let's use singular in both cases and add a period at the end of the sentence:

```
a Standard Gamepad button=] or represents an [=additional gamepad button=].
```

I'm not sure if respec will link the definition correctly if we don't use "Additional gamepad buttons"; if not, we can add a data-lt attribute to the <dfn> tag below:

```
<dfn data-lt="additional gamepad button">Additional gamepad buttons</dfn>
```

index.html
```diff
@@ -878,8 +883,15 @@ <h3>
           <li>Initialize |buttons| to be an empty [=list=].
           </li>
           <li>[=list/For each=] |buttonIndex:long| of [=the range=] from 0 to
-          |buttonsSize| − 1, [=list/append=] a [=new=] {{GamepadButton}} to
-          |buttons|.
+          |buttonsSize| − 1:
+            <ol>
+              <li>Determine the |type| of the button from the provided {{GamepadButtonType}}
+              array, defaulting to null if not specified.
```
nondebug:
```
`null`
```

index.html
```diff
@@ -878,8 +883,15 @@ <h3>
...
+              </li>
+              <li>[=list/Append=] a [=new=] {{GamepadButton}} to |buttons|
```
nondebug: This should reference the |type| variable we defined above:

```
<li>Let |button:GamepadButton| be a [=new=] {{GamepadButton}} with its {{GamepadButton/type}} attribute initialized to |type|.
</li>
<li>[=list/Append=] |button| to |buttons|.
</li>
```

index.html
```diff
@@ -878,8 +883,15 @@ <h3>
           <li>Initialize |buttons| to be an empty [=list=].
           </li>
           <li>[=list/For each=] |buttonIndex:long| of [=the range=] from 0 to
-          |buttonsSize| − 1, [=list/append=] a [=new=] {{GamepadButton}} to
-          |buttons|.
+          |buttonsSize| − 1:
+            <ol>
+              <li>Determine the |type| of the button from the provided {{GamepadButtonType}}
```
nondebug: > Determine the |type| of the button from the provided {{GamepadButtonType}} array

This should be phrased "Let |type| be..." so it's clear we're defining a variable. Also, it's unclear where the GamepadButtonType array came from or how it's being used. If it's provided then it should be passed into this algorithm and given a name, for instance: "To initialize buttons for a gamepad with an array |types| of GamepadButtonType values"

xingri: @nondebug  thanks for the review and detailed guide. Updated this PR to reflect the feedbacks.

xingri:
@marcoscaceres could you please review the changes and let me know if you need anything further.

index.html
```diff
@@ -820,8 +820,13 @@ <h3>
           <li>[=list/For each=] |rawInputIndex:long| of [=the range=] from 0 to
           |inputCount| − 1:
             <ol>
-              <li>If the the gamepad button at index |rawInputIndex|
-              [=represents a Standard Gamepad button=]:
+              <li>Let |button| be the button at index |rawInputIndex|.
```
nondebug: Assigning the button to a variable |button| seems awkward. |button| doesn't have a defined type, it's intended to represent a specific physical button under some device-defined button ordering. Let's reword this to avoid assigning it to anything.

```
<li>Let |type| be the result of determining if the button at index |rawInputIndex|
[=represents a Standard Gamepad button=] or...
</li>
```

index.html
```diff
@@ -878,8 +883,14 @@ <h3>
           <li>Initialize |buttons| to be an empty [=list=].
           </li>
           <li>[=list/For each=] |buttonIndex:long| of [=the range=] from 0 to
-          |buttonsSize| − 1, [=list/append=] a [=new=] {{GamepadButton}} to
-          |buttons|.
+          |buttonsSize| − 1:
+            <ol>
+              <li>Let |button:GamepadButton| be a [=new=] {{GamepadButton}} with its
+              {{GamepadButton/type}} attribute initialized to |type|.
```
nondebug: We're initializing button.type but the remaining GamepadButton attributes are left uninitialized. Let's specify that button.value is initialized to zero and button.pressed and button.touched are initialized to false.

index.html
xingri: @nondebug  Thanks Matt for the review. Just updated the PR.

index.html
xingri: @nondebug  Thanks Matt for the review. Just updated the PR.

index.html
```diff
@@ -878,8 +881,17 @@ <h3>
...
           <li>[=list/For each=] |buttonIndex:long| of [=the range=] from 0 to
-          |buttonsSize| − 1, [=list/append=] a [=new=] {{GamepadButton}} to
-          |buttons|.
+          |buttonsSize| − 1:
+            <ol>
+              <li>Let |button:GamepadButton| be a [=new=] {{GamepadButton}} with its
+              {{GamepadButton/type}} attribute initialized to |type|,
+              button.value initialized to 0,
+              button.pressed initialized to false, and
+              button.touched initialized to false.
```
nondebug: Let's use the same button.attr syntax for all four attributes and reorder them to match the order in the GamepadButton interface definition:

```
              <li>Let |button:GamepadButton| be a [=new=] {{GamepadButton}} with
              |button|.{{GamepadButton/pressed}} initialized to `false`,
              |button|.{{GamepadButton/touched}} initialized to `false`,
              |button|.{{GamepadButton/type}} initialized to |type|, and
              |button|.{{GamepadButton/value}} initialized to 0.
              </li>
```

xingri: @nondebug  Thanks Matt. Just updated.

xingri:
@marcoscaceres , @nondebug  As we discussed today, I have removed the table in the section 12.1.
Please review this PR and let me know your feedback.

index.html
```diff
@@ -2035,6 +2055,60 @@ <h2>
...
+	    {{GamepadButtonType}} enumeration.
+          </p>
+          <p>
+            This enum defines the set of possible button types.
+          </p>
+          <pre class="idl">
+            enum GamepadButtonType {
+              "standard",
+              "non-standard",
+              "trackpad",
```
```suggestion
              "non-standard",
              "standard",
              "trackpad",
```

index.html
```diff
@@ -2035,6 +2055,60 @@ <h2>
           Visual representation of a [=Standard Gamepad=] layout.
         </figcaption>
       </figure>
+      <section>
+        <h3>
+          <dfn data-lt="additional gamepad button">Additional gamepad buttons</dfn>
```
```suggestion
          <dfn>Additional gamepad buttons</dfn>
```

PR #201, opened 2024-04-27
Add sequenceId slot and fix effect promises

### LINKED ISSUE #200: GamepadHapticActuator  promises issues
There appear to be a couple of issues with the steps of `playEffect()` and `reset()` of the [`GamepadHapticActuator Interface `](https://w3c.github.io/gamepad/#dom-gamepadhapticactuator):

`playEffect`:
- "completed" promise can be wrong one if called multiple time: a second call will set the promise to null and queue a "preempted" task, but then immediately set the promise to a new one. The "completed" task for the first call would then run and resolve the second promise. Solution: like in `reset`, check for promise equality. 
- 8.2 checks for promise in parallel steps: unnecessary and reading the internal slot should be done in a task. 

`reset`:
- step 4.4. resolves a promise right on the parallel steps: it should do this inside a task. Also, it is conditional on step 4, so it appears that calling `reset()` when there is no running effect will never resolve the returned promise. Perhaps it should be rejected immediately in that case?
- step 4.3.1 reads slot from parallel step: move it below to inside the task. 

Also it's not clear to me how to compute "are still the same" for two promises, so the spec could be explicit about it and add a generational counter internal slot, that would be passed around to parallel steps by cloning it to a variable and back to queued tasks for comparison with the internal slot.

nondebug:
Closes #200

Implementation commitment:

 * [ ] [WebKit](https://bugs.webkit.org/show_bug.cgi?id=277838)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/pull/201.html" title="Last updated on Jul 10, 2025, 11:07 PM UTC (14dfa9c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/201/c430e73...14dfa9c.html" title="Last updated on Jul 10, 2025, 11:07 PM UTC (14dfa9c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1122,6 +1122,19 @@
             playing.
           </td>
         </tr>
+        <tr>
+          <td>
+            <dfn data-dfn-for="GamepadHapticActuator">[[\sequenceId]]</dfn>
+          </td>
+          <td>
+            0
+          </td>
+          <td>
+            An {{unsigned long}} incremented for each call to
+            {{GamepadHapticActuator/playEffect}} or
+            {{GamepadHapticActuator/reset}}.
+          </td>
+        </tr>
       </table>
       <dl data-dfn-for="GamepadHapticActuator">
         <dt>
@@ -1165,6 +1178,8 @@
             `"hidden"`, return [=a promise rejected with=] an
             "{{InvalidStateError}}" {{DOMException}}.
             </li>
+            <li>Increment [=this=].{{GamepadHapticActuator/[[sequenceId]]}}.
+            </li>
             <li>If [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}}
             is not `null`:
               <ol>
@@ -1185,8 +1200,11 @@
             type=] |type|, return [=a promise rejected with=] reason
             {{NotSupportedError}}.
             </li>
-            <li>Let {{GamepadHapticActuator/[[playingEffectPromise]]}} be [=a
-            new promise=].
+            <li>Let [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}}
+            be [=a new promise=].
+            </li>
+            <li>Let |effectSequenceId:unsigned long| be
+            [=this=].{{GamepadHapticActuator/[[sequenceId]]}}.
             </li>
             <li>Let |playEffectTimestamp:DOMHighResTimestamp| be the [=current
             high resolution time=] given the |document|'s [=relevant global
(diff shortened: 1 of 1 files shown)


marcoscaceres:
@gterzian, do you want to check this also?

marcoscaceres:
Will need to check if the original issue is observable in any browser or if there was code already preventing this.  Might be good if we can come up with a test to confirm (seems it might be possible to do so).

marcoscaceres:
@nondebug I'm still not sure what happens if the identifiers don't match... feels like there should be some kind of assert or something somewhere to assuring nothing is getting out of sync.

index.html
```diff
@@ -1197,29 +1215,29 @@ <h2>
...
-                    is `null`, abort these steps.
-                    </li>
-                    <li>[=Resolve=]
-                    [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}}
-                    with {{GamepadHapticsResult/"complete"}}.
-                    </li>
-                    <li>Set
-                    [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}}
-                    to `null`.
+                    <li>If [=this=].{{GamepadHapticActuator/[[sequenceId]]}} is
```
gterzian, line 1638: You could consider replacing the conditional nesting with an early return(example [usage](https://html.spec.whatwg.org/#dom-trees:parser-inserted-flag)).

index.html
```diff
@@ -1239,37 +1257,57 @@ <h2>
...
+                [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}} to
+                `null`.
+                </li>
+                <li>[=Queue a global task=] on the [=relevant global object=]
+                of [=this=] using the [=gamepad task source=] to [=resolve=]
+                |effectPromise| with {{GamepadHapticsResult/"preempted"}}.
+                </li>
+              </ol>
+            </li>
+            <li>Let [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}}
```
gterzian, line 1696: consider adding a step asserting nullness of the promise at this point(for clarity).

index.html
```diff
@@ -1239,37 +1257,57 @@ <h2>
...
-                    are still the same, set
-                    [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}}
-                    to `null`.
-                    </li>
-                    <li>[=Queue a global task=] on the [=relevant global
-                    object=] of [=this=] using the [=gamepad task source=] to
-                    [=resolve=] |effectPromise| with
-                    {{GamepadHapticsResult/"preempted"}}.
+                    <li>If [=this=].{{GamepadHapticActuator/[[sequenceId]]}} is
+                    |resetSequenceId|:
```
gterzian: same re early return.

index.html
```diff
@@ -1185,8 +1200,11 @@ <h2>
             type=] |type|, return [=a promise rejected with=] reason
             {{NotSupportedError}}.
             </li>
-            <li>Let {{GamepadHapticActuator/[[playingEffectPromise]]}} be [=a
-            new promise=].
+            <li>Let [=this=].{{GamepadHapticActuator/[[playingEffectPromise]]}}
```
gterzian, line 1619: same here on asserting nullness of promise.

gterzian (approved):
LGTM, with a few nits to consider.

marcoscaceres (commented):
Adding @gterzian's suggestions would be great. Can review again after those updates.

index.html
gterzian: On second thought, not sure if an early return makes sense, given we always still need to return the promise at the end(we'd end up with two return statements).

index.html
nondebug: Done

index.html

index.html
nondebug: Done

index.html
nondebug: Done

nondebug:
@gterzian Thanks for the suggestions!

@marcoscaceres Please take another look

marcoscaceres (approved):
approving, but we should add a test so I can check WebKit.

marcoscaceres:
Filed bug https://bugs.webkit.org/show_bug.cgi?id=277838

PR #202, opened 2024-04-27
Add vendorId, productId, name attributes

### LINKED ISSUE #199: Vendor ID, product ID, name attributes
Gamepad identification relies on assumptions about browser-specific behaviors that are not defined as part of the spec. Gamepad API only exposes one device identifier for gamepads, the [Gamepad.id](https://w3c.github.io/gamepad/#dom-gamepad-id) string attribute. The content of this attribute is explicitly left unspecified. Each browser engine uses a different naming scheme.

Chromium typically includes the product name string and 16-bit hexadecimal values for the vendor and product ID. If Gamepad.mapping is "standard" then Chromium also includes a "STANDARD GAMEPAD" tag:

> Wireless Controller (STANDARD GAMEPAD Vendor: 054c Product: 09cc)

Firefox includes the product name string and hexadecimal vendor/product ID values:

> 054c-09cc-Wireless Controller

Safari only includes the product name, which may be rewritten by the OS. If the gamepad is a [GCExtendedGamepad](https://developer.apple.com/documentation/gamecontroller/gcextendedgamepad) (roughly equivalent to Standard Gamepad) then it includes an "Extended Gamepad" tag:

> DUALSHOCK4 Wireless Controller Extended Gamepad

Currently, applications that want to make use of the product name or vendor/product IDs must parse them out of the Gamepad.id string. Gamepad API should provide these identifiers as separate attributes on the Gamepad object and specify their values so they can be used consistently across different browser engines:

```
partial interface Gamepad {
  // The vendor ID of this gamepad, or null if the gamepad does not have a vendor ID.
  readonly attribute unsigned long? vendorId;
  // The product ID of this gamepad, or null if the gamepad does not have a product ID.
  readonly attribute unsigned long? productId;
  // A non-empty human-readable name representing this gamepad.
  readonly attribute DOMString name;
};
```

**Vendor & product ID attributes**

Vendor IDs are allocated by the [USB-IF](https://www.usb.org/developers) and [Bluetooth SIG](https://www.bluetooth.com/specifications/assigned-numbers/). Product IDs are assigned by the manufacturer. These values are typically used to programmatically identify a device, for instance when the operating system selects a device driver.

Web applications use these IDs to identify the gamepad model and provide a more accurate representation of the gamepad and its inputs. For instance, an application may use alternate button icons that match the labels on the gamepad's buttons.

Vendor ID and product ID should be optional attributes. Typically only USB and Bluetooth devices have assigned vendor IDs; Gamepad API should not assume gamepads use a particular transport or protocol. Also, even for devices that have vendor/product IDs, those IDs may not be accessible when accessing the device through the platform's gamepad API. User agents may choose not to expose these IDs in some or all contexts in order to mitigate fingerprinting.

**Name attribute**

The name attribute is intended for use only as a human-facing identifier. Since it is intended to be readable by humans, it should never be empty or null. Typically the value should match the name the manufacturer assigned to the gamepad, but if there is no manufacturer-assigned name (or it cannot be accessed) then the user agent should create a suitable name.

The content of the name attribute cannot be fully specified because it relies on platform behaviors. For instance, on some platforms the device name reported through the platform API only includes the product name (example: "Wireless Controller") while on some platforms it may also include the manufacturer name (example: "Sony Interactive Entertainment Wireless Controller"). The browser is not in control of how these names are generated and cannot assume the name provided by the platform matches the manufacturer-assigned name.

The spec should make it clear that the Gamepad.name attribute is not a stable identifier and is only guaranteed to stay the same as long as the gamepad remains connected. The name may change over time due to browser updates, OS updates, or device firmware updates.

nondebug:
Closes #199

 * [ ] Modified Web platform tests - we will get tests for free. 

Implementation commitment:

 * [ ] [WebKit](https://bugs.webkit.org/show_bug.cgi?id=273981)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/pull/202.html" title="Last updated on Jul 10, 2025, 11:05 PM UTC (54359eb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/202/c430e73...54359eb.html" title="Last updated on Jul 10, 2025, 11:05 PM UTC (54359eb)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -374,7 +374,7 @@
           </p>
           <p class="note">
             The {{Gamepad/name}} is intended only as a user-facing identifier.
-            It is not a stable identifier and may change over time due to
+            It is not a stable identifier and might change over time due to
             updates to the device firmware, browser, or operating system.
           </p>
         </dd>
--- a/index.html
+++ b/index.html
@@ -147,6 +147,9 @@
         [Exposed=Window]
         interface Gamepad {
           readonly attribute DOMString id;
+          readonly attribute unsigned short? vendorId;
+          readonly attribute unsigned short? productId;
+          readonly attribute DOMString name;
           readonly attribute long index;
           readonly attribute boolean connected;
           readonly attribute DOMHighResTimeStamp timestamp;
@@ -332,6 +335,47 @@
             `idProduct` values. Unique identifiers like serial numbers or
             Bluetooth device addresses MUST NOT be included in the
             {{Gamepad/id}} string.
+          </p>
+        </dd>
+        <dt>
+          <dfn>vendorId</dfn> attribute
+        </dt>
+        <dd>
+          <p>
+            The 16-bit identifier assigned to the device vendor by the USB
+            Implementers Forum or Bluetooth Special Interest Group, or `null`
+            if the gamepad has no vendor identifier or it is not available.
+          </p>
+          <p>
+            The [=user agent=] MAY set {{Gamepad/vendorId}} and
+            {{Gamepad/productId}} to `null` to mitigate the risk of exposing
+            device identifiers. If one of {{Gamepad/vendorId}} or
+            {{Gamepad/productId}} is `null` then both MUST be `null`.
+          </p>
+        </dd>
+        <dt>
+          <dfn>productId</dfn> attribute
+        </dt>
+        <dd>
+          The product identifier assigned by the device vendor, or `null` if
+          the gamepad has no product identifier or it is not available.
+        </dd>
+        <dt>
+          <dfn>name</dfn> attribute
+        </dt>
+        <dd>
+          <p>
+            A human-readable identifier for the gamepad. The exact content of
+            the {{Gamepad/name}} string is left unspecified. The [=user agent=]
+            MAY use a product name string provided by the device firmware. The
+            {{Gamepad/name}} MUST NOT be the empty string and MUST NOT include
+            unique identifiers like serial numbers or Bluetooth device
+            addresses.
+          </p>
+          <p class="note">
+            The {{Gamepad/name}} is intended only as a user-facing identifier.
+            It is not a stable identifier and may change over time due to
+            updates to the device firmware, browser, or operating system.
           </p>
         </dd>
         <dt>
@@ -658,6 +702,27 @@
             <ol>
               <li>Initialize |gamepad|'s {{Gamepad/id}} attribute to an
               identification string for the gamepad.
+              </li>
+              <li>If the gamepad has vendor and product identifiers:
+                <ol>
+                  <li>Initialize |gamepad|.{{Gamepad/vendorId}} to the vendor
+                  identifier.
+                  </li>
+                  <li>Initialize |gamepad|.{{Gamepad/productId}} to the product
+                  identifier.
+                  </li>
+                </ol>
+              </li>
+              <li>Otherwise:
+                <ol>
+                  <li>Initialize |gamepad|.{{Gamepad/vendorId}} to `null`.
+                  </li>
+                  <li>Initialize |gamepad|.{{Gamepad/productId}} to `null`.
+                  </li>
+                </ol>
+              </li>
+              <li>Initialize |gamepad|'s {{Gamepad/name}} attribute to a
+              human-readable identifier for the gamepad.
               </li>
               <li>Initialize |gamepad|'s {{Gamepad/index}} attribute to the
               result of [=selecting an unused gamepad index=] for |gamepad|.
--- a/index.html
+++ b/index.html
@@ -703,21 +703,15 @@
               <li>Initialize |gamepad|'s {{Gamepad/id}} attribute to an
               identification string for the gamepad.
               </li>
+              <li>Initialize |gamepad|.{{Gamepad/vendorId}} to `null`</li>
+              <li>Initialize |gamepad|.{{Gamepad/productId}} to `null`.</li>
               <li>If the gamepad has vendor and product identifiers:
                 <ol>
-                  <li>Initialize |gamepad|.{{Gamepad/vendorId}} to the vendor
+                  <li>Set |gamepad|.{{Gamepad/vendorId}} to the vendor
                   identifier.
                   </li>
-                  <li>Initialize |gamepad|.{{Gamepad/productId}} to the product
+                  <li>Set |gamepad|.{{Gamepad/productId}} to the product
                   identifier.
-                  </li>
-                </ol>
-              </li>
-              <li>Otherwise:
-                <ol>
-                  <li>Initialize |gamepad|.{{Gamepad/vendorId}} to `null`.
-                  </li>
-                  <li>Initialize |gamepad|.{{Gamepad/productId}} to `null`.
                   </li>
                 </ol>
               </li>


index.html
```diff
@@ -334,6 +337,47 @@ <h2>
...
+            A human-readable identifier for the gamepad. The exact content of
+            the {{Gamepad/name}} string is left unspecified. The [=user agent=]
+            MAY use a product name string provided by the device firmware. The
+            {{Gamepad/name}} MUST NOT be the empty string and MUST NOT include
+            unique identifiers like serial numbers or Bluetooth device
+            addresses.
+          </p>
+          <p class="note">
+            The {{Gamepad/name}} is intended only as a user-facing identifier.
+            It is not a stable identifier and may change over time due to
```
marcoscaceres: Just avoiding RFC2119 confusion here. 

```suggestion
            It is not a stable identifier and might change over time due to
```

index.html
```diff
@@ -659,6 +703,27 @@ <h3>
...
+                <ol>
+                  <li>Initialize |gamepad|.{{Gamepad/vendorId}} to `null`.
+                  </li>
+                  <li>Initialize |gamepad|.{{Gamepad/productId}} to `null`.
+                  </li>
+                </ol>
+              </li>
+              <li>Initialize |gamepad|'s {{Gamepad/name}} attribute to a
+              human-readable identifier for the gamepad.
+              </li>
```
marcoscaceres: Maybe just: 

```suggestion
              <li>Initialize |gamepad|.{{Gamepad/vendorId}} to `null`</li>
              <li>Initialize |gamepad|.{{Gamepad/productId}} to `null`.</li>
              <li>If the gamepad has vendor and product identifiers:
                <ol>
                  <li>Set |gamepad|.{{Gamepad/vendorId}} to the vendor
                  identifier.
                  </li>
                  <li>Set |gamepad|.{{Gamepad/productId}} to the product
                  identifier.
                  </li>
                </ol>
              </li>
              <li>Initialize |gamepad|'s {{Gamepad/name}} attribute to a
              human-readable identifier for the gamepad.
              </li>
```

marcoscaceres:
Filed webkit bug https://bugs.webkit.org/show_bug.cgi?id=273981

marcoscaceres (approved):
Just some editorial stuff... we are supportive of adding this on the WebKit side.

index.html
```diff
@@ -334,6 +337,47 @@ <h2>
...
+        <dt>
+          <dfn>name</dfn> attribute
+        </dt>
+        <dd>
+          <p>
+            A human-readable identifier for the gamepad. The exact content of
+            the {{Gamepad/name}} string is left unspecified. The [=user agent=]
+            MAY use a product name string provided by the device firmware. The
+            {{Gamepad/name}} MUST NOT be the empty string and MUST NOT include
+            unique identifiers like serial numbers or Bluetooth device
```
marcoscaceres, line 644: Maybe make it more clear that this means the vendorId and productId?

nondebug: I don't consider vendor/product IDs unique identifiers since they typically have the same values for all instances of the same product. The language here was intended to stop implementers from doing something like appending the USB serial number string descriptor to the name attribute as a way to provide a unique identifier. We want to give the implementation some leeway to pick a good name but we shouldn't allow this attribute to be used to hold a unique identifier since that would violate our assumptions around fingerprintability.

Including vendor/product IDs shouldn't make the name string any more useful as a unique identifier since we expect these IDs will correlate strongly with the product name, but I'd argue it still doesn't satisfy the spec because vendor and product IDs aren't really human-readable.

marcoscaceres: ok, yeah, that makes sense. I was more thinking about discouraging implementers from putting weird hex numbers etc. that would then end up being presented to users (which would be meaningless for most users). The description does make it clear that that this is supposed to be an end-user facing string.

marcoscaceres:
@saschanaz, can you confirm that you and folks on the Gecko side are ok with this addition? We are supportive of it on the WebKit side.

index.html
```diff
@@ -147,6 +147,9 @@ <h2>
         [Exposed=Window]
         interface Gamepad {
           readonly attribute DOMString id;
+          readonly attribute unsigned short? vendorId;
```
marcoscaceres, line 398: Should we maybe make this a DOMString to keep the hex representation?

index.html
```diff
@@ -147,6 +147,9 @@ <h2>
...
+          readonly attribute unsigned short? productId;
```
marcoscaceres, line 399: Same as above? otherwise the developer will need to convert this to hex, which can be confusing.

marcoscaceres: We should also provide some guidance here... right now, for instance, Firefox and Chrome are different in their IDs with respect to the "0":

Chrome is:
"DualSense Wireless Controller (STANDARD GAMEPAD Vendor: 054c Product: 0ce6)"
Firefox is:
"54c-ce6-DualSense Wireless Controller"

It might make them easier to identify?

marcoscaceres: Hmm... I guess it's just as easy to do:

```JS
gamepad.productId?.toString(16);
```

marcoscaceres: Actually, yeah... forget it. Let's do unsigned short is fine.

saschanaz:
So this is "fine" as we already have `.id` that exposes the same info, but it's still a fingerprinting vector. Why do we need vendor/product/name at all here? Is it for some feature detection that existing methods do not cover well?

nondebug:
> Why do we need vendor/product/name at all here? Is it for some feature detection that existing methods do not cover well?

Vendor/product IDs enable applications to identify the gamepad model which can be useful when trying to show icons that match the labels on the gamepad's buttons. Currently, applications use Gamepad.id for this but it's not a good identifier because the content of ID string is explicitly not specified and there's no guarantee that the value will remain stable over time. The vendor/product IDs come from the device firmware so they're unlikely to change except by firmware update.

It's also used for feature detection when deducing which "extra" buttons are available. We're trying to solve that use case with a GamepadButton.type attribute for identifying buttons that aren't part of the Standard Gamepad: #196

A human-readable name is important because there may be multiple gamepads connected and the application may need to ask the user to identify a connected gamepad. Gamepad.id can also be used for this but it typically includes other information like vendor/product IDs that makes it less human-readable.

Mostly we want to add these attributes so that developers can eliminate browser-specific logic for parsing this information out of the underspecified Gamepad.id string.

saschanaz:
So three use cases:

1. Show matching icon for each controller
2. Feature detection for buttons
3. Identify each controller

1 may be fun but not sure it's important enough. 2 should be solved by #196. 3 sounds like a workaround rather than actual direct use case of this, how can I identify each controller by name if I have 4 Xbox 360 controllers connected? Should the application instead vibrate each controller for example so that I can know which is which?

It's not adding a new fingerprinting vector, but this still eases use of unnecessarily exposed device specific info that I'd rather discuss removing `id` to resist fingerprinting and add missing features that this might solve as #196 does. (i.e. I think two feature-equivalent gamepads should be perceived equivalent to Gamepad API even if those are different products. Like how we don't allow differentiating a Logitech mouse and Microsoft mouse.)

nondebug:
> 1 may be fun but not sure it's important enough.

Displaying the correct button labels is important for gaming services where players are expected to bring their own gamepad. It's confusing when the game tells you to press the "X button" but you're not sure if it means Xbox X button (buttons[2]), PlayStation Cross button (buttons[0]) or Switch X button (buttons[3]). The gamepad itself doesn't know what the correct labels are so applications need a way to decide which labels to use.

> how can I identify each controller by name if I have 4 Xbox 360 controllers connected?

I think it's not necessary for the human-readable identifier to distinguish between devices that can't be distinguished by a human. If the controllers are identical then (in my opinion) it's fine to use the same identifier.

Sometimes the product name is identical but the device itself may provide additional device-specific distinguishing information. For instance, it's possible to query a Switch Joy-Con's color. As specified in this PR it would be acceptable for a user agent to add the color information to the name string as an additional distinguishing detail. However, most gamepads don't provide a programmatic way to distinguish between cosmetic variations of the same product.

On consoles this problem is solved by assigning each connected gamepad to a slot and using light indicators to distinguish slots. There's a separate proposal #143 to allow applications to set the light indicator state which could be used to distinguish otherwise identical gamepads. This approach is limited by the reality that not all gamepads have light indicators, and those that do only have a limited number of distinct configurations for representing slots.

> Should the application instead vibrate each controller for example so that I can know which is which?

Applications shouldn't rely on vibration since not all gamepads support vibration. Even gamepads that support the feature may give the user a way to disable it, for instance with a physical on-off switch on the gamepad. We can't assume that sending a vibration command to a gamepad will actually cause it to vibrate in a way the user will notice, but vibration could be useful in conjunction with other signals.

> I'd rather discuss removing id to resist fingerprinting and add missing features that this might solve

I don't think removing Gamepad.id is feasible because too many applications already rely on it.

As a fingerprinting mitigation I think it would be reasonable for Gamepad API to offer tiered access to gamepads so that Gamepad.id is safe by default. Perhaps by default a site receives only generic information about connected gamepads similar to how PointerEvents provides almost no information about the connected pointing device. Only the standard buttons/axes are exposed and capabilities like vibration are disabled to resist fingerprinting. After granting a permission, the site would be allowed to access all information about the gamepad, including device IDs. I expect we will need to add a permission model anyway since we want to add support for more powerful gamepad capabilities that wouldn't be appropriate without stronger user consent.

> I think two feature-equivalent gamepads should be perceived equivalent to Gamepad API even if those are different products.

I disagree mainly because of the button labels problem. Other capabilities can be automatically detected, but displaying the correct button labels requires device identification because the labels are never provided by the device itself.

If we want to hide device IDs from the application then the user agent needs to provide the necessary information to select the correct button labels. I expect the fingerprinting risk of exposing labeling information is roughly equivalent to exposing the vendor/product/name since we expect these values to be highly correlated. If the API is responsible for providing this information then it increases the burden on the user agent proportional to the number of supported gamepads. If the fingerprinting risk is roughly equivalent then I think we should save ourselves the effort and let applications developers build tables of button labels for the devices they care about.

marcoscaceres:
Just adding that I agree with @nondebug's assessment. On the WebKit side, we see the user interaction requirement as sufficient to mitigate the fingerprinting concerns. At the same time, if users agents want to go beyond that for the productId and vendorId (and sometimes not provide them, or provide generalized identifiers) to provide further anonymity, they are certainly free to do so.  

There is limited opportunity for two or more sites to collude to identify an individual because of both the user activation are requirement, and because the set of controllers types out there is expansive, but not super unique per individual (and there's other sadly more effective means to fingerprint). 

As such, we might want to make that clear in the spec (that the user agent can lie if it needs to or the user wants it to).

marcoscaceres (approved):

saschanaz:
> Displaying the correct button labels is important for gaming services where players are expected to bring their own gamepad. It's confusing when the game tells you to press the "X button" but you're not sure if it means Xbox X button (buttons[2]), PlayStation Cross button (buttons[0]) or Switch X button (buttons[3]). The gamepad itself doesn't know what the correct labels are so applications need a way to decide which labels to use.

I agree that's a real problem! Still I don't think exposing vendor/product ID is a right solution, as the caller would need a database to tell which ID has which mapping. Fortunately there are a few well-defined button mappings, so wonder we could have a proper API to expose that info, perhaps with an enum. (As you noted later this still would require the mapping from browser side, which I would say okay and is just easier to use.)

> I think it's not necessary for the human-readable identifier to distinguish between devices that can't be distinguished by a human. If the controllers are identical then (in my opinion) it's fine to use the same identifier.

But that means having IDs does not really solve the core problem? If it only helps when the user is having different type of controllers, I'm not sure how that would be helpful most of the time.

Again, this is "okay" as it does not expose additional info, but I still does not understand why we want to solve problems with IDs that does not expose any useful device info as-is.

nondebug:
> Fortunately there are a few well-defined button mappings, so wonder we could have a proper API to expose that info, perhaps with an enum.

Button mapping isn't equivalent to deducing button labels. Console gamepads often have the same layout but with different labels. It's important for applications to be able to draw the correct labels even for gamepads with the same button layout and mapping. Suppose the game is trying to tell the player which button opens an options menu and wants the player to press the button mapped to buttons[9] ("right button in center cluster"). It's not enough for the game to know that this gamepad uses a PlayStation layout because the labels will differ depending on if it's a PS3, PS4, or PS5 controller. On [DualShock 3](https://en.wikipedia.org/wiki/DualShock#/media/File:DualShock_3.jpg) this button is labeled "Start", on [DualShock 4](https://en.wikipedia.org/wiki/DualShock#/media/File:DualShock_4.jpg) it's "Options", and on [DualSense](https://en.wikipedia.org/wiki/DualShock#/media/File:Playstation_DualSense_Controller.png) it's a hamburger icon. For the game to draw the correct label, the enum at minimum will need to distinguish between these possibilities. When a new PlayStation gamepad is released, it's likely that the spec would need to be updated with a new enum value.

We should try to avoid building this sort of dependency into the spec where the spec needs to be updated for each new device we want to support. Ideally it should be possible to connect your gamepad to your browser on the day the gamepad is released and have it work correctly without needing to wait for a spec change and a browser update. If we make it the browser's responsibility to maintain button label information then I expect there will be a delay before new gamepads are supported and less popular gamepads may never be supported.

Also, I think we should try to avoid referencing specific vendors or trademarks in the spec. If we added a enum with different values for "well-defined button mappings" then I expect the enum will include many trademarked names by necessity.

> But that means having IDs does not really solve the core problem?

The core problem (as I see it) is that applications are relying on the underspecified Gamepad.id attribute for programmatic identification of gamepad models. Applications already maintain mappings of Gamepad.id -> button labels. This makes it difficult to modify the Gamepad.id value for any currently-supported gamepad because it will break applications that rely on that gamepad having a stable Gamepad.id value. The purpose of the name attribute isn't to guarantee distinguishability, it's to provide the same level of human-facing distinguishability currently provided by Gamepad.id without requiring identifier stability.

> If it only helps when the user is having different type of controllers, I'm not sure how that would be helpful most of the time.

Do we need to consider whether it's helpful? I agree it's typically not necessary for users to identify a gamepad because usually there's only one connected gamepad, or all the gamepads are the same product model. However, the current API provides a human-facing identifier and application developers sometimes find it useful to present this identifier to users. I think we should assume that a human-facing identifier is helpful to application developers and empower browser implementations to provide the best possible identifier.

marcoscaceres:
Agree with @nondebug on this and on the WebKit side. We think it makes sense to split these up as it's something developers rely on. 

At the same time, there is certainly scope for improving the privacy generally here - but I think we should take that up separately.

saschanaz:
Okay, Mozilla is neutral here, given the ID has been being exposed and pages already can extract each ID component by some heuristic. We would be happier without the need to sniff based on ID but also acknowledge that it would be hard to deal with use cases given the hardware reality.

marcoscaceres:
@nondebug raised a case where pre-release hardware was exposed through these values. I guess a thing to be mindful of would be if the browser gives users control to not expose these. There's a balance to be struck here, as the product and vendor id would have still been exposed in the id string.

pedropaulosuzuki:
> > If it only helps when the user is having different type of controllers, I'm not sure how that would be helpful most of the time.
> 
> Do we need to consider whether it's helpful? I agree it's typically not necessary for users to identify a gamepad because usually there's only one connected gamepad, or all the gamepads are the same product model. However, the current API provides a human-facing identifier and application developers sometimes find it useful to present this identifier to users. I think we should assume that a human-facing identifier is helpful to application developers and empower browser implementations to provide the best possible identifier.

Just adding some context as a game developer that also plays games: we often have one controller of each model (one Dualshock 4, one Dualshock 5, one Steam Controller, one Xbox 360 controller, one Xbox Series controller, one Switch Pro controller, and so on) to be able to test all possible controller configurations for our games, and as we have many controllers, having more than one of each model is overkill and generally a waste of money. Thus, for us, it's very common when playing a local co-op game to use different controller models, so I wouldn't say this is as niche as it seems. Also, players sometimes have one Dualshock control for playing Playstation emulated games and one XInput controller for the rest of PC games, and that's also more common that one would think. I can also think of all the players that had a single Dualshock 4, upgraded to a Dualshock 5, but still kept the older controller for co-op. I could go on and on, but I think you get the point.

PR #207, opened 2024-05-10
Require system focus for gamepad user gesture

### LINKED ISSUE #206: Require focus for initial gamepad user gesture
Gamepad API uses the "gamepad user gesture" as a signal that the user consents to exposing gamepad information. This makes it possible to unintentionally expose gamepad information when interacting with gamepads on another site.

The spec should be amended to require that the window is focused when the gesture is received so that access is only granted for the focused tab.

nondebug:
Closes #206

Implementation commitment:

 * [ ] [WebKit](https://bugs.webkit.org/show_bug.cgi?id=277836)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/pull/207.html" title="Last updated on May 14, 2026, 11:38 PM UTC (5ba33c8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/207/d8c7780...5ba33c8.html" title="Last updated on May 14, 2026, 11:38 PM UTC (5ba33c8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1683,8 +1683,8 @@
           The {{Navigator/getGamepads()}} method steps are:
         </p>
         <ol>
-          <li>Let |doc| be the [=current global object=]'s [=associated
-          `Document`=].
+          <li>Let |doc:Document?| be the [=current global object=]'s
+          [=associated `Document`=].
           </li>
           <li>If |doc| is `null` or |doc| is not [=Document/fully active=],
           then return an empty [=list=].
@@ -1699,6 +1699,8 @@
           <li>Let |now:DOMHighResTimeStamp| be the [=current high resolution
           time=].
           </li>
+          <li>Let |gamepads:sequence&lt;Gamepad?&gt;| be an empty [=list=].
+          </li>
           <li>[=list/For each=] |gamepad:Gamepad| of
           [=this=].{{Navigator/[[gamepads]]}}:
             <ol>
@@ -1710,6 +1712,8 @@
                   <li>Set |gamepad|.{{Gamepad/[[timestamp]]}} to |now|.
                   </li>
                 </ol>
+              </li>
+              <li>[=list/Append=] |gamepad| to |gamepads|.
               </li>
             </ol>
           </li>
--- a/index.html
+++ b/index.html
@@ -511,7 +511,9 @@
           object=]'s {{Navigator}} object.
           </li>
           <li>If |navigator|.{{Navigator/[[hasGamepadGesture]]}} is `false` and
-          |gamepad| [=contains a gamepad user gesture=]:
+          |navigator|'s [=top-level traversable=] has [=top-level
+          traversable/system focus=] and |gamepad| [=contains a gamepad user
+          gesture=]:
             <ol>
               <li>Set |navigator|.{{Navigator/[[hasGamepadGesture]]}} to
               `true`.
--- a/index.html
+++ b/index.html
@@ -510,10 +510,13 @@
           <li>Let |navigator:Navigator| be |gamepad|'s [=relevant global
           object=]'s {{Navigator}} object.
           </li>
+          <li>Let |document:Document?| be |gamepad|'s [=relevant global
+          object=]'s [=associated `Document`=]; otherwise `null`.
+          </li>
           <li>If |navigator|.{{Navigator/[[hasGamepadGesture]]}} is `false` and
-          |navigator|'s [=top-level traversable=] has [=top-level
-          traversable/system focus=] and |gamepad| [=contains a gamepad user
-          gesture=]:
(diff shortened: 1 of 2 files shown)


marcoscaceres:
This seems reasonable. I need to check if WebKit already does this or not.

nondebug:
> This seems reasonable. I need to check if WebKit already does this or not.

I think WebKit on macOS already requires focus even for normal gamepad inputs because it uses Game Controller API which requires focus.

marcoscaceres:
We might actually want this:
https://html.spec.whatwg.org/multipage/interaction.html#fully-active-descendant-of-a-top-level-traversable-with-user-attention

* Fully active
* visibly requirement 
* and system system focus!

marcoscaceres (approved):
This looks ok to me.

index.html
```diff
@@ -801,8 +801,13 @@ <h3>
...
           object=]'s {{Navigator}} object.
           </li>
+          <li>Let |document:Document?| be |gamepad|'s [=relevant global
+          object=]'s [=associated `Document`=]; otherwise `null`.
+          </li>
           <li>If |navigator|.{{Navigator/[[hasGamepadGesture]]}} is `false` and
-          |gamepad| [=contains a gamepad user gesture=]:
+          |document| is a [=Document/fully active descendant of a top-level
+          traversable with user attention=] and |gamepad| [=contains a gamepad
+          user gesture=]:
```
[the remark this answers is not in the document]
marcoscaceres: Ok, yeah... maybe this check should also happen earlier. If you don't have a document anyway, then none of the above really matters, right?

index.html
```diff
@@ -801,8 +801,13 @@ <h3>
           <li>Let |navigator:Navigator| be |gamepad|'s [=relevant global
           object=]'s {{Navigator}} object.
           </li>
+          <li>Let |document:Document?| be |gamepad|'s [=relevant global
+          object=]'s [=associated `Document`=]; otherwise `null`.
+          </li>
           <li>If |navigator|.{{Navigator/[[hasGamepadGesture]]}} is `false` and
-          |gamepad| [=contains a gamepad user gesture=]:
+          |document| is a [=Document/fully active descendant of a top-level
+          traversable with user attention=] and |gamepad| [=contains a gamepad
```
gabrielsanbrito, line 811: Should we create an alias for "focused"?

"A [=Document=] |document| is focused if |document| is a [=Document/fully active descendant of a top-level traversable with user attention=] and |gamepad| [=contains a gamepad user gesture=]:"?

index.html
gabrielsanbrito, line 2533: Thinking about our current event model and the upcoming [Input Events](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/GamepadEventDrivenInputAPI/explainer.md), should we also only dispatch events if the tab is focused?

index.html
marcoscaceres: I don’t think we should do that, to be honest. If we need the concept of focus, we should link out to HTML (which defines it).

index.html
```diff
@@ -801,8 +801,13 @@ <h3>
           <li>Let |navigator:Navigator| be |gamepad|'s [=relevant global
           object=]'s {{Navigator}} object.
           </li>
+          <li>Let |document:Document?| be |gamepad|'s [=relevant global
+          object=]'s [=associated `Document`=]; otherwise `null`.
```
marcoscaceres, line 805: If the document is `null` then that's exceptional... InvalidStateError?

index.html
```diff
@@ -2112,6 +2113,8 @@ <h3>
           </li>
           <li>Let |gamepads:sequence&lt;Gamepad?&gt;| be an empty [=list=].
           </li>
+          <li>Let |gamepads:sequence&lt;Gamepad?&gt;| be an empty [=list=].
+          </li>
```
marcoscaceres: Duplicate step — `Let |gamepads:sequence<Gamepad?>| be an empty [=list=].` appears twice. Suggest removing the second one:

```suggestion
```

index.html
nondebug: I think it doesn't make sense to throw here because we're inside "When the system receives new button or axis input values", if we throw an error there's nothing to catch it.

marcoscaceres: Ah, good point. Yes, it should just return gracefully.

gabrielsanbrito (approved):
LGTM

index.html
gabrielsanbrito: Connection/disconnection events shouldn't require focus to be dispatched. If `Navigator.hasGamepadGesture` is `true`, this means that the user has at least once interacted with a gamepad while the page was focused. It is ok to fire connection/disconnection events in this case.

PR #217, opened 2024-10-10
Make eventInitDict param optional in GamepadEvent constructor

### LINKED ISSUE #216: eventInitDict parameter in GamepadEvent constructor should be optional
The [GamepadEvent constructor](https://w3c.github.io/gamepad/#dom-gamepadevent-constructor) is defined with two required parameters:

```
constructor(DOMString type, GamepadEventInit eventInitDict);
```

This differs from the signature in implementations which define the second parameter as optional ([blink](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/gamepad/gamepad_event.idl;drc=047c7dc4ee1ce908d7fea38ca063fa2f80f92c77), [Gecko](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/Modules/gamepad/GamepadEvent.idl), [WebKit](https://searchfox.org/mozilla-central/source/dom/webidl/GamepadEvent.webidl)):

```
  constructor(DOMString type, optional GamepadEventInit eventInitDict = {}); // blink
  constructor(DOMString type, optional GamepadEventInit eventInitDict = {});  // Gecko
  constructor([AtomString] DOMString type, optional GamepadEventInit eventInitDict);  // WebKit
```

nondebug:
Closes #216

Implementation commitment:

 * [X] WebKit
 * [X] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/pull/217.html" title="Last updated on Feb 13, 2026, 12:49 AM UTC (ffd83a2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/217/0186e68...ffd83a2.html" title="Last updated on Feb 13, 2026, 12:49 AM UTC (ffd83a2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1771,7 +1771,7 @@
         </h3>
         <pre class="idl" data-cite="DOM">
         dictionary GamepadEventInit : EventInit {
-          required Gamepad gamepad;
+          Gamepad? gamepad = null;
         };
       </pre>
         <dl data-dfn-for="GamepadEventInit">
--- a/index.html
+++ b/index.html
@@ -1752,7 +1752,7 @@
         [Exposed=Window]
 
         interface GamepadEvent: Event {
-          constructor(DOMString type, GamepadEventInit eventInitDict);
+          constructor(DOMString type, optional GamepadEventInit eventInitDict);
           [SameObject] readonly attribute Gamepad gamepad;
         };
       </pre>
--- a/index.html
+++ b/index.html
@@ -1752,7 +1752,7 @@
         [Exposed=Window]
 
         interface GamepadEvent: Event {
-          constructor(DOMString type, optional GamepadEventInit eventInitDict);
+          constructor(DOMString type, optional GamepadEventInit eventInitDict = {});
           [SameObject] readonly attribute Gamepad gamepad;
         };
       </pre>


index.html
```diff
@@ -1752,7 +1752,7 @@ <h2>
         [Exposed=Window]
 
         interface GamepadEvent: Event {
-          constructor(DOMString type, GamepadEventInit eventInitDict);
+          constructor(DOMString type, optional GamepadEventInit eventInitDict = {});
```
marcoscaceres, line 2168: We probably need to remove the `required` from GamepadEventInit's `gamepad`.

nondebug: Blink, Gecko, WebKit agree that GamepadEventInit's `gamepad` attribute is optional and nullable, let's update the spec to match.

```
  Gamepad? gamepad;  // blink
  Gamepad? gamepad = null;  // Gecko
  Gamepad? gamepad = null;  // WebKit
```

marcoscaceres: Sounds good.

tcl3:
Should the type of the `gamepad` attribute also be nullable, since it can now return `null`?

marcoscaceres:
@tcl3, yes, good catch. I think the attributes are required to mirror the value they are initialized with.

weinig:
Other than matching current implementations, is there a reason/use case for a null `gamepad` on GamepadEvent. In WebKit at least, the engine will never create a `GamepadEvent` with a null `Gamepad` (other than in this case where the user constructs it from JS). Is there reason to think there is a compat issue with fixing engines to require it?

marcoscaceres (approved):

PR #220, opened 2025-03-19
Add a model section

nondebug:
Closes #???

Implementation commitment:

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/pull/220.html" title="Last updated on Jun 12, 2025, 10:24 PM UTC (422ecb0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/220/fbde104...422ecb0.html" title="Last updated on Jun 12, 2025, 10:24 PM UTC (422ecb0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -136,6 +136,164 @@
         sensing, depth sensing, video analysis, gesture recognition, and so on.
       </p>
     </section>
+    <section>
+      <h2>
+        Model
+      </h2>
+      <p>
+        A <dfn>gamepad</dfn> is a collection of input controls (buttons,
+        triggers, joysticks, thumbsticks, touch surfaces) and output controls
+        (haptic actuators). A [=gamepad=] is <dfn>available</dfn> if the [=user
+        agent=] can read the current state of its input controls. A [=gamepad=]
+        that is not [=available=] is <dfn>unavailable</dfn>. The collection of
+        input and output controls for a [=gamepad=] cannot change while the
+        [=gamepad=] is [=available=]. Each input control has one or more
+        <dfn>input values</dfn> that may update over time.
+      </p>
+      <p>
+        The [=user agent=] is responsible for enumerating [=available=]
+        [=gamepads=], detecting when [=gamepads=] become [=available=] or
+        [=unavailable=], and detecting when input controls have updated input
+        values.
+      </p>
+      <p>
+        A [=gamepad=] has a <dfn>gamepad identifier string</dfn>, a
+        human-readable string that identifies the brand or style of
+        [=gamepad=]. The content is decided by the [=user agent=].
+      </p>
+      <p>
+        A [=gamepad=] may be a physical device. If so, it has an <dfn>input
+        control layout</dfn> that describes the position, orientation and type
+        of each input control on the physical device. The [=user agent=] is
+        responsible for recognizing when an [=input control layout=]
+        <dfn data-lt="corresponds with">corresponds with a standard
+        layout</dfn>, meaning the relative positions and orientations of its
+        input controls are close enough to the standard layout that a user can
+        use it interchangeably with other devices that have the same input
+        controls and correspond with the same layout.
+      </p>
+      <p>
+        A [=gamepad=] may have [=axis=] inputs. An <dfn>axis</dfn> is an input
+        control with the following [=input values=]:
+      </p>
+      <ul>
+        <li>A <dfn>logical axis value</dfn>, an integer value that represents
+        the current displacement of the control from a reference position.
+        </li>
+        <li>A <dfn>minimum axis value</dfn>, the smallest in-bounds [=logical
+        axis value=].
+        </li>
+        <li>A <dfn>maximum axis value</dfn>, the largest in-bounds [=logical
+        axis value=].
+        </li>
+      </ul>
+      <p>
+        A [=gamepad=] has an <dfn>axis list</dfn> which is a [=list=]
+        containing all the [=axis=] inputs for the [=gamepad=] in some order
+        determined by the [=user agent=].
+      </p>
+      <p>
+        An input control may be designed to automatically return an [=axis=] to
+        a center position when the user stops interacting with the input
+        control. If so, the [=axis=] has a <dfn>preferred axis state</dfn>. An
+        An [=axis=] with a [=preferred axis state=] may also have a <dfn>center
+        position value</dfn> [=input value=] which is the [=logical axis
+        value=] when the [=axis=] is centered.
+      </p>
+      <p>
+        A [=gamepad=] may have [=button=] inputs. A <dfn>button</dfn> is an
+        input control that can be pressed to activate. A [=gamepad=] has a
+        <dfn>button list</dfn> which is a [=list=] containing all the
+        [=button=] inputs for the [=gamepad=] in some order determined by the
+        [=user agent=].
+      </p>
+      <p>
+        An input control may be designed to automatically return a [=button=]
+        to an unpressed state when the user stops interacting with the input
+        control. If so, the [=button=] has a <dfn>preferred button state</dfn>.
+      </p>
+      <p>
+        A [=button=] may have a digital switch to indicate when the [=button=]
+        is activated. If so, the [=button=] has a <dfn>digital button
+        value</dfn> [=input value=] that is `true` when the [=button=] is
+        activated and `false` otherwise.
+      </p>
+      <p>
+        A [=button=] may have an analog sensor enabling the [=button=] to
+        report the degree to which the [=button=] is activated. If so, the
+        [=button=] has these [=input values=]:
+      </p>
+      <ul>
+        <li>An <dfn>analog button value</dfn>, an integer value that represents
+        the degree of activation.
+        </li>
+        <li>A <dfn>minimum analog value</dfn> representing no activation. The
+        smallest in-bounds [=analog button value=].
+        </li>
+        <li>A <dfn>maximum analog value</dfn> representing full activation. The
+        largest in-bounds [=analog button value=].
+        </li>
+      </ul>
+      <p>
+        A [=button=] may be capable of detecting touch. If so, the [=button=]
+        has a <dfn>button touch value</dfn> [=input value=] that is `true` when
+        the [=button=] is touched and `false` otherwise.
+      </p>
+      <p>
+        A [=gamepad=] may have [=touch surfaces=]. A <dfn>touch surface</dfn>
+        is an input control that provides 2D position data representing points
+        of contact. A [=gamepad=] has a <dfn>touch surface list</dfn> which is
+        a [=list=] containing the [=touch surfaces=] for the [=gamepad=]. The
+        list is ordered such that [=touch surfaces=] closer to the left side of
+        the [=gamepad=] appear closer to the start of the list.
+      </p>
+      <p>
+        A [=touch surface=] has an <dfn>active touch point list</dfn> [=input
+        value=], a [=list=] of zero or more [=touch points=] representing the
+        points of contact currently detected by the sensor, ordered such that
+        earlier [=touch points=] appear closer to the start of the [=list=]. A
+        <dfn>touch point</dfn> represents a single point of contact at a single
+        point in time. A [=touch point=] has a <dfn>touch x coordinate</dfn>
+        representing its position along the user's left-right axis and a
+        <dfn>touch y coordinate</dfn> representing its position along the
+        perpendicular axis.
+      </p>
+      <p>
+        A [=touch surface=] may have surface dimension [=input values=]. The
+        <dfn>surface width</dfn> and <dfn>surface height</dfn> [=input values=]
+        are the dimensions of the [=touch surface=] in the same units as the
+        [=touch x coordinate=] and [=touch y coordinate=]. A [=touch surface=]
+        has both dimension values or neither value.
+      </p>
+      <p>
+        A [=touch point=] may be a new contact point or a continuation of an
+        earlier contact. A [=touch point=] is <dfn>part of an existing active
+        touch point</dfn> if the [=user agent=] identifies that it is a
+        continuation of a [=touch point=] represented by an earlier
+        {{GamepadTouch}}. The <dfn>active touch point id</dfn> for a [=touch
+        point=] is the {{GamepadTouch/touchId}} of the earlier
+        {{GamepadTouch}}.
+      </p>
+      <p>
+        A [=gamepad=] may have [=haptic actuators=]. A <dfn>haptic
+        actuator</dfn> is an output control capable of moving the [=gamepad=]
+        in a way that can be felt by the user. A <dfn>haptic effect</dfn> is
+        the use of one or more [=haptic actuators=] to provide feedback to the
+        user. The [=user agent=] is responsible for commanding [=haptic
+        actuators=] to play and stop [=haptic effects=] on [=available=]
+        [=gamepads=].
+      </p>
+      <p>
+        A [=gamepad=] may have a <dfn>vibration actuator</dfn> which is a
+        [=haptic actuator=] capable of playing a [=haptic effect=] to vibrate
+        the whole [=gamepad=].
+      </p>
+      <p>
+        A [=haptic actuator=] has a [=list=] of <dfn>supported effect
+        types</dfn> containing one or more {{GamepadHapticEffectType}} values,
+        which cannot change while the [=gamepad=] is [=available=].
+      </p>
+    </section>
     <section data-dfn-for="Gamepad" data-link-for="Gamepad">
       <h2>
         <dfn>Gamepad</dfn> interface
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -136,6 +136,164 @@ <h2>
...
+        that is not [=available=] is <dfn>unavailable</dfn>. The collection of
+        input and output controls for a [=gamepad=] cannot change while the
+        [=gamepad=] is [=available=]. Each input control has one or more
+        <dfn>input values</dfn> that may update over time.
+      </p>
+      <p>
+        The [=user agent=] is responsible for enumerating [=available=]
+        [=gamepads=], detecting when [=gamepads=] become [=available=] or
+        [=unavailable=], and detecting when input controls have updated input
+        values.
```
gabrielsanbrito: Maybe make this a bullet list?

index.html
```diff
@@ -136,6 +136,164 @@ <h2>
         sensing, depth sensing, video analysis, gesture recognition, and so on.
       </p>
     </section>
+    <section>
+      <h2>
+        Model
+      </h2>
+      <p>
+        A <dfn>gamepad</dfn> is a collection of input controls (buttons,
```
gabrielsanbrito:
```suggestion
        A <dfn>gamepad</dfn> is a collection of <dfn>input controls</dfn> (buttons,
```
Maybe we should also add a definition for "input controls"?

index.html
```diff
@@ -136,6 +136,164 @@ <h2>
...
+        [=unavailable=], and detecting when input controls have updated input
+        values.
+      </p>
+      <p>
+        A [=gamepad=] has a <dfn>gamepad identifier string</dfn>, a
+        human-readable string that identifies the brand or style of
+        [=gamepad=]. The content is decided by the [=user agent=].
+      </p>
+      <p>
+        A [=gamepad=] may be a physical device. If so, it has an <dfn>input
```
gabrielsanbrito: Can a gamepad be something other than a physical device in the context of this spec? If so, maybe we can exemplify which types of non-physical devices a gamepad object can represent?

index.html
```diff
@@ -136,6 +136,164 @@ <h2>
...
+        control layout</dfn> that describes the position, orientation and type
+        of each input control on the physical device. The [=user agent=] is
+        responsible for recognizing when an [=input control layout=]
+        <dfn data-lt="corresponds with">corresponds with a standard
+        layout</dfn>, meaning the relative positions and orientations of its
+        input controls are close enough to the standard layout that a user can
+        use it interchangeably with other devices that have the same input
+        controls and correspond with the same layout.
```
gabrielsanbrito: I think that this clause is a little confusing. This part to be more specific: "that a user can use it interchangeably with other devices that have the same input controls and correspond with the same layout". 

I think that adding "standard" at the end makes things clearer?
```suggestion
        layout</dfn>, meaning the relative positions and orientations of its
        input controls are close enough to a standard layout that a user can
        use it interchangeably with other devices that have the same input
        controls and correspond with the same standard layout.
```

index.html
```diff
@@ -136,6 +136,164 @@ <h2>
...
+        containing all the [=axis=] inputs for the [=gamepad=] in some order
+        determined by the [=user agent=].
+      </p>
+      <p>
+        An input control may be designed to automatically return an [=axis=] to
+        a center position when the user stops interacting with the input
+        control. If so, the [=axis=] has a <dfn>preferred axis state</dfn>. An
+        An [=axis=] with a [=preferred axis state=] may also have a <dfn>center
+        position value</dfn> [=input value=] which is the [=logical axis
+        value=] when the [=axis=] is centered.
```
gabrielsanbrito: Remove double "An" and maybe we remove the `[=input value=]` reference to make the text more fluid? Or maybe we can have `<dfn>center position input value</dfn>`?
```suggestion
        control. If so, the [=axis=] has a <dfn>preferred axis state</dfn>. An
        [=axis=] with a [=preferred axis state=] may also have a <dfn>center
        position value</dfn> which is the [=logical axis
        value=] when the [=axis=] is centered.
```

gabrielsanbrito: This same feedback also applies to `<dfn>digital button value</dfn>` and `dfn>button touch value</dfn>`.

index.html
```diff
@@ -136,6 +136,164 @@ <h2>
...
+        A [=gamepad=] may have [=touch surfaces=]. A <dfn>touch surface</dfn>
+        is an input control that provides 2D position data representing points
+        of contact. A [=gamepad=] has a <dfn>touch surface list</dfn> which is
+        a [=list=] containing the [=touch surfaces=] for the [=gamepad=]. The
+        list is ordered such that [=touch surfaces=] closer to the left side of
+        the [=gamepad=] appear closer to the start of the list.
+      </p>
+      <p>
+        A [=touch surface=] has an <dfn>active touch point list</dfn> [=input
+        value=], a [=list=] of zero or more [=touch points=] representing the
```
gabrielsanbrito, line 288:
```suggestion
        A [=touch surface=] has an <dfn>active touch point list</dfn> of [=input
        values=], a [=list=] of zero or more [=touch points=] representing the
```

index.html
```diff
@@ -136,6 +136,164 @@ <h2>
...
+      </ul>
+      <p>
+        A [=button=] may be capable of detecting touch. If so, the [=button=]
+        has a <dfn>button touch value</dfn> [=input value=] that is `true` when
+        the [=button=] is touched and `false` otherwise.
+      </p>
+      <p>
+        A [=gamepad=] may have [=touch surfaces=]. A <dfn>touch surface</dfn>
+        is an input control that provides 2D position data representing points
+        of contact. A [=gamepad=] has a <dfn>touch surface list</dfn> which is
```
gabrielsanbrito, line 281: Is it relevant to mention that the we are using a cartesian system to locate these touch points?
```suggestion
        A [=gamepad=] may have [=touch surfaces=]. A <dfn>touch surface</dfn>
        is an input control that provides 2D position data representing points
        of contact in a cartesian coordinate system, where the origin is located in the
        center of the touch surface. A [=gamepad=] has a <dfn>touch surface list</dfn> which is
```

index.html
```diff
@@ -136,6 +136,164 @@ <h2>
...
+        the [=gamepad=] appear closer to the start of the list.
+      </p>
+      <p>
+        A [=touch surface=] has an <dfn>active touch point list</dfn> [=input
+        value=], a [=list=] of zero or more [=touch points=] representing the
+        points of contact currently detected by the sensor, ordered such that
+        earlier [=touch points=] appear closer to the start of the [=list=]. A
+        <dfn>touch point</dfn> represents a single point of contact at a single
+        point in time. A [=touch point=] has a <dfn>touch x coordinate</dfn>
+        representing its position along the user's left-right axis and a
```
gabrielsanbrito: Maybe "horizontal" is more formal?
```suggestion
        representing its position along the user's horizontal axis and a
```

index.html
```diff
@@ -136,6 +136,164 @@ <h2>
...
+        <dfn>touch y coordinate</dfn> representing its position along the
+        perpendicular axis.
+      </p>
+      <p>
+        A [=touch surface=] may have surface dimension [=input values=]. The
```
gabrielsanbrito, line 300: Are the surface dimensions considered `input values`? According to our definition, `input values` are associated with `input controls`. Maybe we could just say that dimensions are just "values"?

index.html
```diff
@@ -136,6 +136,164 @@ <h2>
...
+        has both dimension values or neither value.
+      </p>
+      <p>
+        A [=touch point=] may be a new contact point or a continuation of an
+        earlier contact. A [=touch point=] is <dfn>part of an existing active
+        touch point</dfn> if the [=user agent=] identifies that it is a
+        continuation of a [=touch point=] represented by an earlier
+        {{GamepadTouch}}. The <dfn>active touch point id</dfn> for a [=touch
+        point=] is the {{GamepadTouch/touchId}} of the earlier
+        {{GamepadTouch}}.
```
gabrielsanbrito:
```suggestion
        {{GamepadTouch}}. The <dfn>active touch point id</dfn> for a [=touch
        point=] that is [=part of an existing active touch point=] is the
        {{GamepadTouch/touchId}} of the earlier {{GamepadTouch}}.
```
Or maybe:
```suggestion
        {{GamepadTouch}}. In this case, the [=touch point=]'s {{GamepadTouch/touchId}} is the 
        {{GamepadTouch/touchId}} of the earlier {{GamepadTouch}}.
```

index.html
```diff
@@ -136,6 +136,164 @@ <h2>
...
+      </p>
+      <p>
+        A [=gamepad=] may have [=haptic actuators=]. A <dfn>haptic
+        actuator</dfn> is an output control capable of moving the [=gamepad=]
+        in a way that can be felt by the user. A <dfn>haptic effect</dfn> is
+        the use of one or more [=haptic actuators=] to provide feedback to the
+        user. The [=user agent=] is responsible for commanding [=haptic
+        actuators=] to play and stop [=haptic effects=] on [=available=]
+        [=gamepads=].
```
gabrielsanbrito: Maybe this way?
```suggestion
        A [=gamepad=] may have [=haptic actuators=]. A <dfn>haptic
        actuator</dfn> is an output control capable of moving the [=gamepad=]
        in a way that can be felt by the user. One or more [=Haptic actuators=]
        can be used to generate <dfn>haptic effects</dfn> and provide feedback to the
        user. The [=user agent=] is responsible for commanding [=haptic
        actuators=] to play and stop [=haptic effects=] on [=available=]
        [=gamepads=].
```

index.html
```diff
@@ -136,6 +136,164 @@ <h2>
         sensing, depth sensing, video analysis, gesture recognition, and so on.
       </p>
     </section>
+    <section>
+      <h2>
+        Model
+      </h2>
```
gabrielsanbrito, line 142: Maybe we can create a subdivide this section in:
- Model 
  - input controls
    - axes
    - buttons
    - touch 
  - output controls

In the Model's section caput we can define the gamepad's properties (basically the first 4 paragraphs) and then we can go into the controls' subsections. WDYT of something like this?
```
//Model
A gamepad is a collection of input controls...
[...]
... and correspond with the same layout.

A [=gamepad=] may have input controls:
- Axis inputs
- button inputs
- touch surfaces

A [=gamepad=] may have output controls like haptic actuators.

// // Input controls
[...]
// // Output controls
```

gabrielsanbrito (commented):
@nondebug left a couple of comments. I think that the model section looks good!

index.html
nondebug: Good idea. I thought of a few more things that should go here (reading the input values and controlling the output controls)

index.html
nondebug: Yes, let's add definitions for both input controls and output controls.

index.html
nondebug: Here I'm trying to define [=gamepad=] so that the input layout is optional. In the Automation section I'm planning to define "simulated gamepad" as a [=gamepad=] so it can be passed to algorithms that accept a [=gamepad=]. Gamepads created by WebDriver won't have an input layout so I want to note here that it's optional.

> Can a gamepad be something other than a physical device

Yes, it can be a virtual device or a WebDriver simulated gamepad. It's a bit of a tangent and it's not relevant here so I think I'll try to rewrite this to remove the "physical" distinction and focus on the fact that the input layout is optional.

The issue is that, as defined, an input layout only exists if the gamepad is a physical device but the browser doesn't actually have a way to distinguish physical devices from virtual devices. A virtual device might be designed to behave indistinguishably from a physical device, or the difference between real and physical gamepads might be obscured by the system input API. The spec should allow the browser to use heuristics to assume the input layout given hints from device identifiers or the system input API.

index.html
nondebug: Yes, this is confusing. I tried refactoring it a bit to simplify.

index.html
nondebug: It needs to be clear which values are [=input values=] and which are just properties of the input control. I prefer tagging them with the cross-reference as it leaves no room for ambiguity. I'll try rewording to make this flow better while keeping [=input value=].

index.html
nondebug: I think this doesn't work. The user agent detects when input values are updated, but if there are no input values (zero touch points) then what is the user agent detecting? That's why I defined it as a single input value containing a list.

index.html
nondebug: Yes, let's note it's a Cartesian system since we only include normalization steps for Cartesian coordinates.

I don't think we can say anything about the [=touch surface=] origin, that's determined by the surface's logical representation which can have arbitrary bounds. The origin is moved to the center by normalization.

index.html
nondebug: When we describe the thumbstick axes we use both Horizontal and Left/Right, let's do the same here to minimize ambiguity.

I think there's another issue here, aligning to the horizontal axis only makes sense if the touchpad is on the top/bottom/front/back of the gamepad. If it's on the left or right side then the left/right axis is perpendicular to the surface of the touchpad. To make this generic we should define the coordinate system for any orientation.

index.html
nondebug: I went back and forth on this. In general I want to divide gamepad properties into "read when the gamepad is first connected and never changes" and "read each time inputs are read from the gamepad". We don't expect touch surface dimensions to change since they're supposed to represent the dimensions of the hardware touch sensor. But we never require that they *can't* change, so it seems safer to treat them as if they could change. What if it's a virtual gamepad that lets the user resize the touch surface area?

index.html
nondebug: I went with the first one, the second doesn't work because a [=touch point=] doesn't have a {{GamepadTouch/touchId}} attribute

index.html
nondebug: Yes, "the use of" was awkward and this is more fluid. Modified to emphasize that a single effect can be composed of vibrations from multiple actuators:

> A [=haptic actuator=] can be used to generate a **haptic effect** that provides feedback to the user. Vibrations from multiple actuators are combined to generate more complex haptic effects.

index.html
nondebug: Adding more structure SGTM.

index.html
gabrielsanbrito: Makes sense. It's ok by me to omit the origin detail

index.html
```diff
@@ -136,6 +136,204 @@ <h2>
...
+        The [=user agent=] SHOULD consider a layout to correspond with a standard layout if its [=input controls=] have approximately the same relative positions and orientations as [=input controls=] described in the standard layout.
+      </p>
+      <p>
+        The [=user agent=] typically cannot directly inspect the [=input control layout=] for a [=gamepad=] and MAY use heuristics to decide the layout.
+        The [=user agent=] SHOULD consider the device identifiers when deciding whether a [=gamepad=] [=corresponds with a standard layout=].
+        If the system assigns a label to each [=input control=] and the labels imply a particular layout then the [=user agent=] SHOULD consider the [=gamepad=] to have that layout.
+        When there is a standard model and an accessible model with the same [=input controls=], the [=user agent=] SHOULD consider the accessible model to have the same [=input control layout=] as the standard model.
```
gabrielsanbrito, line 186: Quick question. What would be an "accessible" model?

index.html
```diff
@@ -136,6 +136,204 @@ <h2>
...
+      </p>
+      <p>
+        A [=touch surface=] has an <dfn>active touch point list</dfn> [=input
+        value=], a [=list=] of zero or more [=touch points=] representing the
+        points of contact currently detected by the sensor. A <dfn>touch
+        point</dfn> represents a single point of contact at a single point in
+        time. A [=touch point=] has a <dfn>touch x coordinate</dfn>
+        and a <dfn>touch y coordinate</dfn> representing the position in the [=touch surface=]'s coordinate system.
+        If a [=touch surface=] is on the top, bottom, front, or back side of the [=gamepad=] then the [=touch x coordinate=] is measured along the left-right axis, otherwise it is measured along the top-bottom axis.
+        The [=touch y coordinate=] is measured along the perpendicular axis.
```
gabrielsanbrito, line 294: Should we mention anything about the orientation of the y-axis?

index.html
gabrielsanbrito: Got it. Sounds good then.

gabrielsanbrito (approved):
Sorry for the long delay. I gave it a last pass and I think that the model section looks good. Just added 2 comments, which are basically questions. Thanks for making it so detailed!

index.html
nondebug: I was thinking of gamepads like Xbox Adaptive Controller and PlayStation Access where the manufacturer's intent is a swap-in replacement for a gamepad with a standard layout. We should handle these accessibility-focused devices as if they the same input control layout as the non-accessible version of the gamepad (Xbox Wireless Controller or DualSense) even though the physical layout of of the input controls doesn't match the standard layout.

index.html
nondebug: We should, since the current spec is ambiguous about which way is +Y. For other axes we define this in the normalization steps, let's do the same for touch coordinates.

PR #224, opened 2025-06-12
Integrate with WebDriver and WebDriver BiDi for automation

nondebug:
Closes #???

Implementation commitment:

 * [ ] [WebKit](https://bugs.webkit.org/show_bug.cgi?id=309849)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/pull/224.html" title="Last updated on Mar 12, 2026, 11:43 PM UTC (02c8a13)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/gamepad/224/fbde104...02c8a13.html" title="Last updated on Mar 12, 2026, 11:43 PM UTC (02c8a13)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -136,6 +136,164 @@
         sensing, depth sensing, video analysis, gesture recognition, and so on.
       </p>
     </section>
+    <section>
+      <h2>
+        Model
+      </h2>
+      <p>
+        A <dfn>gamepad</dfn> is a collection of input controls (buttons,
+        triggers, joysticks, thumbsticks, touch surfaces) and output controls
+        (haptic actuators). A [=gamepad=] is <dfn>available</dfn> if the [=user
+        agent=] can read the current state of its input controls. A [=gamepad=]
+        that is not [=available=] is <dfn>unavailable</dfn>. The collection of
+        input and output controls for a [=gamepad=] cannot change while the
+        [=gamepad=] is [=available=]. Each input control has one or more
+        <dfn>input values</dfn> that may update over time.
+      </p>
+      <p>
+        The [=user agent=] is responsible for enumerating [=available=]
+        [=gamepads=], detecting when [=gamepads=] become [=available=] or
+        [=unavailable=], and detecting when input controls have updated input
+        values.
+      </p>
+      <p>
+        A [=gamepad=] has a <dfn>gamepad identifier string</dfn>, a
+        human-readable string that identifies the brand or style of
+        [=gamepad=]. The content is decided by the [=user agent=].
+      </p>
+      <p>
+        A [=gamepad=] may be a physical device. If so, it has an <dfn>input
+        control layout</dfn> that describes the position, orientation and type
+        of each input control on the physical device. The [=user agent=] is
+        responsible for recognizing when an [=input control layout=]
+        <dfn data-lt="corresponds with">corresponds with a standard
+        layout</dfn>, meaning the relative positions and orientations of its
+        input controls are close enough to the standard layout that a user can
+        use it interchangeably with other devices that have the same input
+        controls and correspond with the same layout.
+      </p>
+      <p>
+        A [=gamepad=] may have [=axis=] inputs. An <dfn>axis</dfn> is an input
+        control with the following [=input values=]:
+      </p>
+      <ul>
+        <li>A <dfn>logical axis value</dfn>, an integer value that represents
+        the current displacement of the control from a reference position.
+        </li>
+        <li>A <dfn>minimum axis value</dfn>, the smallest in-bounds [=logical
+        axis value=].
+        </li>
+        <li>A <dfn>maximum axis value</dfn>, the largest in-bounds [=logical
+        axis value=].
+        </li>
+      </ul>
+      <p>
+        A [=gamepad=] has an <dfn>axis list</dfn> which is a [=list=]
+        containing all the [=axis=] inputs for the [=gamepad=] in some order
+        determined by the [=user agent=].
+      </p>
+      <p>
+        An input control may be designed to automatically return an [=axis=] to
+        a center position when the user stops interacting with the input
+        control. If so, the [=axis=] has a <dfn>preferred axis state</dfn>. An
+        An [=axis=] with a [=preferred axis state=] may also have a <dfn>center
+        position value</dfn> [=input value=] which is the [=logical axis
+        value=] when the [=axis=] is centered.
+      </p>
+      <p>
+        A [=gamepad=] may have [=button=] inputs. A <dfn>button</dfn> is an
+        input control that can be pressed to activate. A [=gamepad=] has a
+        <dfn>button list</dfn> which is a [=list=] containing all the
+        [=button=] inputs for the [=gamepad=] in some order determined by the
+        [=user agent=].
+      </p>
+      <p>
+        An input control may be designed to automatically return a [=button=]
+        to an unpressed state when the user stops interacting with the input
+        control. If so, the [=button=] has a <dfn>preferred button state</dfn>.
+      </p>
+      <p>
+        A [=button=] may have a digital switch to indicate when the [=button=]
+        is activated. If so, the [=button=] has a <dfn>digital button
+        value</dfn> [=input value=] that is `true` when the [=button=] is
+        activated and `false` otherwise.
+      </p>
+      <p>
+        A [=button=] may have an analog sensor enabling the [=button=] to
+        report the degree to which the [=button=] is activated. If so, the
+        [=button=] has these [=input values=]:
+      </p>
+      <ul>
+        <li>An <dfn>analog button value</dfn>, an integer value that represents
+        the degree of activation.
+        </li>
+        <li>A <dfn>minimum analog value</dfn> representing no activation. The
+        smallest in-bounds [=analog button value=].
+        </li>
+        <li>A <dfn>maximum analog value</dfn> representing full activation. The
+        largest in-bounds [=analog button value=].
+        </li>
+      </ul>
+      <p>
+        A [=button=] may be capable of detecting touch. If so, the [=button=]
+        has a <dfn>button touch value</dfn> [=input value=] that is `true` when
+        the [=button=] is touched and `false` otherwise.
+      </p>
+      <p>
+        A [=gamepad=] may have [=touch surfaces=]. A <dfn>touch surface</dfn>
+        is an input control that provides 2D position data representing points
+        of contact. A [=gamepad=] has a <dfn>touch surface list</dfn> which is
+        a [=list=] containing the [=touch surfaces=] for the [=gamepad=]. The
+        list is ordered such that [=touch surfaces=] closer to the left side of
+        the [=gamepad=] appear closer to the start of the list.
+      </p>
+      <p>
+        A [=touch surface=] has an <dfn>active touch point list</dfn> [=input
+        value=], a [=list=] of zero or more [=touch points=] representing the
+        points of contact currently detected by the sensor, ordered such that
+        earlier [=touch points=] appear closer to the start of the [=list=]. A
+        <dfn>touch point</dfn> represents a single point of contact at a single
+        point in time. A [=touch point=] has a <dfn>touch x coordinate</dfn>
+        representing its position along the user's left-right axis and a
+        <dfn>touch y coordinate</dfn> representing its position along the
+        perpendicular axis.
+      </p>
+      <p>
+        A [=touch surface=] may have surface dimension [=input values=]. The
+        <dfn>surface width</dfn> and <dfn>surface height</dfn> [=input values=]
+        are the dimensions of the [=touch surface=] in the same units as the
+        [=touch x coordinate=] and [=touch y coordinate=]. A [=touch surface=]
+        has both dimension values or neither value.
+      </p>
+      <p>
+        A [=touch point=] may be a new contact point or a continuation of an
+        earlier contact. A [=touch point=] is <dfn>part of an existing active
+        touch point</dfn> if the [=user agent=] identifies that it is a
+        continuation of a [=touch point=] represented by an earlier
+        {{GamepadTouch}}. The <dfn>active touch point id</dfn> for a [=touch
+        point=] is the {{GamepadTouch/touchId}} of the earlier
+        {{GamepadTouch}}.
+      </p>
+      <p>
+        A [=gamepad=] may have [=haptic actuators=]. A <dfn>haptic
+        actuator</dfn> is an output control capable of moving the [=gamepad=]
+        in a way that can be felt by the user. A <dfn>haptic effect</dfn> is
+        the use of one or more [=haptic actuators=] to provide feedback to the
+        user. The [=user agent=] is responsible for commanding [=haptic
+        actuators=] to play and stop [=haptic effects=] on [=available=]
+        [=gamepads=].
+      </p>
+      <p>
+        A [=gamepad=] may have a <dfn>vibration actuator</dfn> which is a
+        [=haptic actuator=] capable of playing a [=haptic effect=] to vibrate
+        the whole [=gamepad=].
+      </p>
+      <p>
+        A [=haptic actuator=] has a [=list=] of <dfn>supported effect
+        types</dfn> containing one or more {{GamepadHapticEffectType}} values,
+        which cannot change while the [=gamepad=] is [=available=].
+      </p>
+    </section>
     <section data-dfn-for="Gamepad" data-link-for="Gamepad">
       <h2>
         <dfn>Gamepad</dfn> interface
--- a/index.html
+++ b/index.html
@@ -2889,7 +2889,7 @@
               <li>
                 Let |simulatedGamepads| be the [=simulated gamepad map=].
               </li>
-              <li>
+              <li data-cite="webcrypto">
                 Let |deviceId:string| be the result of calling [=generate a random UUID=].
               </li>
               <li>
--- a/index.html
+++ b/index.html
@@ -175,7 +175,7 @@
           Input control layout
         </h3>
       <p>
-        A [=gamepad=] may have an <dfn>input control layout</dfn> that describes the position, orientation and type of each input control on the [=gamepad=].
+        A [=gamepad=] may have an <dfn>input control layout</dfn> that describes the position, orientation and type of each [=input control=] on the [=gamepad=].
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2406,6 +2608,913 @@ <h2>
...
+                  <li>
+                    If |params.vibration| [=list/is empty=], |gamepad| has no [=vibration actuator=].
+                    Otherwise, it has a [=vibration actuator=] with the [=supported effect types=] listed in |params|.|vibration|.
+                  </li>
+                </ul>
+              </li>
+              <li>
+                Let |simulatedGamepads| be the [=simulated gamepad map=].
+              </li>
+              <li>
```
```suggestion
              <li data-cite="webcrypto">
```

index.html
```diff
@@ -136,6 +136,204 @@ <h2>
...
+          [=Input values=] can update at any time.
+          The [=user agent=] is responsible for detecting when [=input values=] have updated and SHOULD try to minimize the delay between the update and when the updated values are read.
+        </p>
+        <p>
+          Reading an [=input value=] returns its <dfn>logical value</dfn>, an unscaled numerical representation of the current state.
+          An [=input value=] also has a <dfn>logical minimum</dfn> and <dfn>logical maximum</dfn> which define the minimum and maximum in-bounds [=logical values=].
+        </p>
+        <p>
+          An [=input value=] may have an associated <dfn>HID usage identifier</dfn>, a 32-bit value that identifies the type of data represented by the input value.
+          A HID usage does not precisely describe the [=input control layout=], but by convention many [=gamepads=] with similar layouts use similar usages.
```
sadym-chromium, line 204: > by convention many [=gamepads=] with similar layouts use similar usages

It would be great to link a resource with the list of conventional HID usage identifiers. Even if it is not strictly standardizes, it would be helpful to have such a list for reference.

index.html
```diff
@@ -136,6 +136,204 @@ <h2>
...
+      </p>
+      <p>
+        A [=touch surface=] has an <dfn>active touch point list</dfn> [=input
+        value=], a [=list=] of zero or more [=touch points=] representing the
+        points of contact currently detected by the sensor. A <dfn>touch
+        point</dfn> represents a single point of contact at a single point in
+        time. A [=touch point=] has a <dfn>touch x coordinate</dfn>
+        and a <dfn>touch y coordinate</dfn> representing the position in the [=touch surface=]'s coordinate system.
+        If a [=touch surface=] is on the top, bottom, front, or back side of the [=gamepad=] then the [=touch x coordinate=] is measured along the left-right axis, otherwise it is measured along the top-bottom axis.
+        The [=touch y coordinate=] is measured along the perpendicular axis.
```
sadym-chromium, line 289: NIT: should we define the specific orientation of the the perpendicular axis?

index.html
```diff
@@ -587,6 +795,16 @@ <h3>
...
         </ol>
+        <p>
+          To <dfn>normalize a logical axis value</dfn> given |value|, |minimum|, and |maximum|:
+        </p>
+        <ol>
+          <li>
+            Return 2 (|value| − |minimum|) / (|maximum| − |minimum|) − 1.
+          </li>
+          <li>
+          </li>
```
sadym-chromium:
```suggestion
```

index.html
```diff
@@ -587,6 +795,16 @@ <h3>
             </ol>
           </li>
         </ol>
+        <p>
+          To <dfn>normalize a logical axis value</dfn> given |value|, |minimum|, and |maximum|:
+        </p>
+        <ol>
+          <li>
+            Return 2 (|value| − |minimum|) / (|maximum| − |minimum|) − 1.
```
sadym-chromium, line 803: I would expect the normalized output of input with min `1`, max `5` and val `3` to be `1/2`:
```suggestion
            Return (|value| − |minimum|) / (|maximum| − |minimum|).
```

index.html
```diff
@@ -587,6 +795,16 @@ <h3>
...
+          </li>
+          <li>
+          </li>
+        </ol>
         <p>
           To <dfn>map and normalize axes</dfn> for |gamepad:Gamepad|, run the
           following steps:
```
sadym-chromium, line 808: NIT: The definition below now can be formulated in strict terms introduced previously: `input values` #dfn-input-values and `axis list` #dfn-axis-list
>
```
<li>Let |axisValues:list| be a [=list=] of {{unsigned long}} values
representing the most recent logical axis input values for each axis
input of the device represented by |gamepad|.
</li>```

Comment (sadym-chromium) [reply index.html:808]: Also `mappedIndex` is not used, so probably it can be omitted.

File: index.html [hunk 7]
```diff
@@ -1361,6 +1585,8 @@ <h2>
                 <li>[=Issue a haptic effect=] to the actuator with |type|,
                 |params|, and the |playEffectTimestamp|.
                 </li>
+                <li>If |this|'s [=gamepad=] is a [=simulated gamepad=], run the steps to [=vibrate a simulated gamepad=] for |this|'s [=gamepad=] with |params|.
+                </li>
```
sadym-chromium, line 1587: I assume we should do only one of these steps:
```suggestion
                <li>If |this|'s [=gamepad=] is a [=simulated gamepad=], run the steps to [=vibrate a simulated gamepad=] for |this|'s [=gamepad=] with |params|.
                </li>
                <li>Otherwise, [=issue a haptic effect=] to the actuator with |type|,
                |params|, and the |playEffectTimestamp|.
                </li>
```

index.html
```diff
@@ -2255,7 +2483,7 @@ <h3 id="event-gamepadconnected">
         The <dfn class="event">gamepadconnected</dfn> event
       </h3>
       <p>
-        When a gamepad becomes available on the system, run the following
+        <dfn>When a gamepad becomes available on the system</dfn>, run the following
```
sadym-chromium, line 2484: I believe it's more correct to split this into a trigger and a process.
```suggestion
        When a [=gamepad=] <var>gamepad</var> becomes available on the system,
        [=emit `gamepadconnected` event=] for <var>gamepad</var>.
      </p>
      <p>
        To <dfn>emit `gamepadconnected` event</dfn> for given [=gamepad=] <var>gamepad</var>, run the following
```

sadym-chromium: Also I believe the `[=current global object=]` is ambiguous, especially in case of simulated devices.

index.html
```diff
@@ -2310,7 +2538,7 @@ <h3 id="event-gamepaddisconnected">
         The <dfn class="event">gamepaddisconnected</dfn> event
       </h3>
       <p>
-        When a gamepad becomes unavailable on the system, run the following
+        <dfn>When a gamepad becomes unavailable on the system</dfn>, run the following
```
sadym-chromium, line 2539: The same issue as with `When a gamepad becomes available on the system`.

index.html
```diff
@@ -2406,6 +2634,1425 @@ <h2>
...
+    <section id='automation'>
+      <h2>
+        Automation
+      </h2>
+      <p>
+        For the purposes of [=user agent=] automation and application testing, this document defines extensions to the [[WebDriver-BiDi]] expecification.
+        It is OPTIONAL for a [=user agent=] to support them.
+      </p>
+      <p>
+        Each [=top-level traversable=] has a <dfn>simulated gamepad map</dfn> which is an [=ordered map=] of [=version 4 UUIDs=] to [=gamepad input sources=].
```
sadym-chromium, line 2644: I think the UUID key it sufficient and we don't need to have it ordered map.
```suggestion
        Each [=top-level traversable=] has a <dfn>simulated gamepad map</dfn> which is a [=map=] of [=version 4 UUID=] and [=gamepad input source=].
```

sadym-chromium, line 2644: Perhaps it's `simulated gamepad`, not `gamepad input source`:
```suggestion
        Each [=top-level traversable=] has a <dfn>simulated gamepad map</dfn> which is an [=ordered map=] of [=version 4 UUIDs=] to [=simulated gamepad=].
```

index.html
```diff
@@ -2406,6 +2634,1425 @@ <h2>
...
+            Types
+          </h4>
+          <section>
+            <h5>
+              The emulation.GamepadBounds Type
+            </h5>
+            <pre class="cddl">
+              emulation.GamepadBounds = {
+                max: float64,
+                ? min: float64,
```
sadym-chromium, line 2668: You can define `.default` in CDDL like this:
```suggestion
                ? min: float64 .default 0.0,
```

index.html
```diff
@@ -2406,6 +2634,1425 @@ <h2>
...
+            </pre>
+          </section>
+          <section>
+            <h5>
+              The emulation.SimulateGamepadParams Type
+            </h5>
+            <pre class="cddl">
+              emulation.SimulateGamepadParams = {
+                type: "gamepad",
+                context: browsingContext.BrowsingContext,
```
sadym-chromium, line 2740: IDK if we can cross-reference CDDL to BiDi spec, so let's use `text` the same way [Bluetooth spec does](https://webbluetoothcg.github.io/web-bluetooth/#issue-194ad560).
```suggestion
                context: text,
```

index.html
sadym-chromium: I believe the UUID version is important only when it is generated:
```suggestion
        Each [=top-level traversable=] has a <dfn>simulated gamepad map</dfn> which is an [=ordered map=] of [=UUID=] to [=gamepad input sources=].
```

index.html
```diff
@@ -2406,6 +2634,1425 @@ <h2>
...
+            </pre>
+            <p>
+              The [=remote end steps=] with command parameters |params| are:
+            </p>
+            <ol>
+              <li>
+                Let |simulatedGamepads| be the [=simulated gamepad map=].
+              </li>
+              <li>
+                If |simulatedGamepads|[|params|.|deviceId|] does not [=map/exist=], abort these steps.
```
sadym-chromium, line 2923: 1. We should return a error if there is no such an simulated device.
2. `params` is a map, so `|params|["<code>deviceId</code>"]` is the correct way to access it's values by keys.
```suggestion
                If |simulatedGamepads|[|params|["<code>deviceId</code>"]] does not [=map/exist=], return [=error=]
                with [=error code=] [=invalid argument=].
```

index.html
```diff
@@ -2406,6 +2634,1425 @@ <h2>
...
+              </li>
+              <li>
+                Let |gamepad| be |simulatedGamepads|[|params|.|deviceId|].
```
sadym-chromium, line 2926:
```suggestion
                Let |gamepad| be |simulatedGamepads|[|params|["<code>deviceId</code>"]].
```

index.html
```diff
@@ -2406,6 +2634,1425 @@ <h2>
...
+              </li>
+              <li>
+                Run the steps [=when a gamepad becomes unavailable on the system=] with |gamepad|.
+              </li>
+              <li>
+                [=map/Remove=] |simulatedGamepads|[|params|.|deviceId|].
```
sadym-chromium, line 2932:
```suggestion
                [=map/Remove=] |simulatedGamepads|[|params|["<code>deviceId</code>"]].
```

index.html
```diff
@@ -2406,6 +2634,1425 @@ <h2>
...
+              The emulation.simulatedDeviceEvent Event
+            </h5>
+            <pre class="cddl">
+              emulation.SimulatedDeviceEvent = {
+                method: "emulation.removeSimulatedDevice",
+                params: emulation.SimulatedDeviceEventParams,
+              }
+            </pre>
+            <p>
+              To get <dfn>the identifier for a simulated gamepad</dfn> |gamepad|:
```
sadym-chromium, line 2952: Optionally, we can also extend the definition of "simulated gamepad" as "Simulated gamepad has an associated [=UUID=] <dfn>simulated gamepad identifier</dfn>".

index.html
```diff
@@ -2406,6 +2634,1425 @@ <h2>
...
+            </ol>
+            <p>
+              To <dfn>vibrate a simulated gamepad</dfn> |gamepad| given a [=navigable=] |navigable| and {{GamepadEffectParameters}} |effectParams|:
+            </p>
+            <ol>
+              <li>
+                Let |deviceId:string| be [=the identifier for a simulated gamepad=] |gamepad|.
+              </li>
+              <li>
+                If |deviceId| is `null`, abort these steps.
```
sadym-chromium, line 2978: Maybe we want to notify the user that the given gamepad is removed by throwing an exception?

index.html
```diff
@@ -2406,6 +2634,1425 @@ <h2>
...
+                Let |eventParams| be a [=map=] matching the `emulation.SimulatedDeviceEventParameters` production with the `deviceId` field set to |deviceId|.
+              </li>
+              <li>
+                Let |vibration| be `false`.
+              </li>
+              <li>
+                Let |channels| be the [=list=] « `"strongMagnitude"`, `"weakMagnitude"`, `"leftTrigger"`, `"rightTrigger"` ».
+              </li>
+              <li>
+                [=list/For each=] |channel| of |channels|:
```
sadym-chromium, line 2993: It means that we ignore all other potential values. IDK if this is correct, and how extensible it is.

index.html
```diff
@@ -2406,6 +2634,1425 @@ <h2>
...
+                </ol>
+              </li>
+              <li>
+                [=map/Set=] |eventParams|[`"vibration"`] to |vibration|.
+              </li>
+              <li>
+                Let |body| be a [=map=] matching the `emulation.SimulatedDeviceEvent` production with the `params` field set to |eventParams|.
+              </li>
+              <li>
+                Let |relatedNavigables| be a [=set=] containing |navigable|.
```
sadym-chromium, line 3015: Note: this means that even if the event was initiated from an iframe, the event will be emitted only on the top-level navigable. I believe this is fine.

index.html
```diff
@@ -2406,6 +2634,1425 @@ <h2>
...
+      <section>
+        <h3>
+          The input module
+        </h3>
+        <section data-cite="infra">
+          <h4>
+            Gamepad Input Source
+          </h4>
+          <p>
+            A <dfn>gamepad input source</dfn> is an [=input source=] that is associated with a [=gamepad=]-like device.
```
sadym-chromium, line 3069: Do we need the `gamepad input source` instance? Cannot it be a `simulated gamepad`?

index.html
```diff
@@ -2406,6 +2634,1425 @@ <h2>
...
+              <dt>
+                type
+              </dt>
+              <dd>
+                must be `"touchEnd"`.
+              </dd>
+            </dl>
+          </section>
+          <p>
+            Extend the dispatch action algorithm table in [=dispatch tick actions=] with the following table:
```
sadym-chromium, line 3643: Please add reference to the WebDriver Classic spec, where `dispatch tick actions` is defined:
```
spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
    text: dispatch tick actions; url: dfn-dispatch-tick-actions
```

index.html
```diff
@@ -2406,6 +2634,1425 @@ <h2>
...
+              Let |bounds| be equal to |axes|[|axis index|].
+            </li>
+            <li>
+              If |bounds| is `undefined`, return [=success=] with data `null`.
+            </li>
+            <li>
+              Let |simulated gamepads| be the [=simulated gamepad map=].
+            </li>
+            <li>
+              If |simulated gamepads|[|device id|] does not [=map/exist=], abort these steps.
```
sadym-chromium, line 3767: Return a error with "invalid argument" or a new error like "no such simulated gamepad"

index.html
```diff
@@ -2406,6 +2634,1425 @@ <h2>
...
+          <h4>
+            Gamepad Input Source
+          </h4>
+          <p>
+            A <dfn>gamepad input source</dfn> is an [=input source=] that is associated with a [=gamepad=]-like device.
+          </p>
+          <p>
+            A [=gamepad input source=] has the following items:
+          </p>
+          <table>
```
sadym-chromium, line 3074: Consider defining the `gamepad input source` (or `simulated gamepad`) as a [=struct=] with [=struct/item=] ([example](https://github.com/w3c/webdriver-bidi/blob/32892f339e3a749e7cf6640fe30f0ef8857a34c4/index.bs#L2955-L2957)). That would allow for referencing the items with the links to it's definitions and search for references ([example](https://www.w3.org/TR/webdriver-bidi/#download-behavior-struct-destination-folder)).

index.html
```diff
@@ -2406,6 +2634,1425 @@ <h2>
...
+              Let |subtype| be the result of getting the property named "type" from |action item|.
+            </li>
+            <li>
+              If |subtype| is not one of the values `"axis"`, `"button"`, `"touchStart"`, `"touchMove"`, `"touchEnd"`, or `"pause"`, return an [=error=] with [=error code=] [=invalid argument=].
+            </li>
+            <li>
+              Let |action| be an [=action object=] constructed with arguments |id|, `"gamepad"`, and |subtype|.
+            </li>
+            <li>
+              If |subtype| is `"pause"`, let |result| be the result of [=trying=] to [=process a pause action=] with arguments |action item| and |action|, and return |result|.
```
sadym-chromium, line 3250: This and other calls can be simplified:
```suggestion
              If |subtype| is `"pause"`, return the result of [=trying=] to [=process a pause action=] |action item| and |action|.
```

index.html
```diff
@@ -2406,6 +2634,1425 @@ <h2>
...
+            </li>
+            <li>
+              If |subtype| is `"touchMove"`, let |result| be the result of [=trying=] to [=process a gamepad touch move action=] with arguments |action item| and |action|, and return |result|.
+            </li>
+            <li>
+              If |subtype| is `"touchEnd"`, let |result| be the result of [=trying=] to [=process a gamepad touch end action=] with arguments |action item| and |action|, and return |result|.
+            </li>
+            <li>
+              Return [=success=] with data |action|.
+            </li>
```
sadym-chromium, line 3269: Should not it be an "invalid argument" error, as the actions is unknown?

index.html
```diff
@@ -2406,6 +2634,1425 @@ <h2>
...
+              `vibration` set to |params|.vibration.
+            </li>
+          </ul>
+        </section>
+        <section>
+          <h4>
+            Actions
+          </h4>
+          <p>
+            To <dfn>process a gamepad action</dfn> given |id| and |action item|:
```
sadym-chromium, line 3237: It looks like this part is not finished yet. Please let me know if I can help with it.

index.html
lauromoura: For a complete list of usage identifiers, there's https://www.usb.org/document-library/hid-usage-tables-16 (a 443-page PDF)?

